PR #6, opened 2020-06-10
Add load account button

msfeldstein:
Currently the wallet only supports creating a new random keypair.  This adds the ability to load a wallet from a secret key.

I will comment in-line on what actually changed since a lot of this is autoformat changes.  I will also submit a pr after this that runs autoformat on everything and adds a pre-commit check so we don't deal with these kinds of changes anymore


src/components/wallet/events/render.tsx
```diff
@@ -1,33 +1,103 @@
-import { h } from '@stencil/core'
-import { has as loHas } from 'lodash-es'
+import { h } from "@stencil/core";
+import { has as loHas } from "lodash-es";
 
 export default function render() {
+  const loggedInContent = () => [
```
msfeldstein, line 5: I pulled out the contents of the body into the loggedIn/OutContent functions to keep things clearer vs putting so much in a ternary operator in the body.

src/components/wallet/events/render.tsx
```diff
@@ -1,33 +1,103 @@
...
+      onClick={() => this.createAccount()}
+    >
+      {this.loading.fund ? <stellar-loader /> : null} Create Account
+    </button>,
+    <button
+      class={this.loading.loadAccount ? "loading" : null}
+      type="button"
+      onClick={(e) => this.loadAccount(e)}
+    >
+      {this.loading.loadAccount ? <stellar-loader /> : null} Load Account
```
msfeldstein, line 60: I added this button

src/components/wallet/wallet.ts
```diff
@@ -1,78 +1,73 @@
...
-  updateAccount = updateAccount
-  depositAsset = depositAsset // NEW
-  withdrawAsset = withdrawAsset // NEW
-  trustAsset = trustAsset
-  makePayment = makePayment
-  copyAddress = copyAddress
-  copySecret = copySecret
-  signOut = signOut
+  createAccount = createAccount;
+  loadAccount = loadAccount;
```
msfeldstein, line 58: Other than this line its all formatting changes

PR #7, opened 2020-06-10
Add formatter and run formatting on all files

msfeldstein:
It helps to have an auto-formatter on a project touched by multiple people so we don't end up with peoples formatting tools fighting each other or inconsistent formatting between commits.

In the [first commit](https://github.com/stellar/stellar-demo-wallet/commit/9f069a2f8521ad4a8e934bc79741c227ef19648e) I added prettier and husky, in order to run prettier on any changed files on commit to make sure files are always formatted properly

In the second commit i run prettier on all files so we get all the formatting changes at once and won't have to deal with it in the future.

### DIFF
--- a/electron/app.js
+++ b/electron/app.js
@@ -1,16 +1,16 @@
-const { menubar } = require('menubar')
-const path = require('path')
+const { menubar } = require("menubar");
+const path = require("path");
 
 const cb = menubar({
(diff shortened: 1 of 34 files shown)


msfeldstein:
I'm going to redo this i didn't pull latest master before doing this 😭

Issue #15, opened 2020-07-06
Instrument SEP24 with logging


msfeldstein:
Use the new logview component to instrument the sep24 functionality to replace the demo client

[Deposit](https://github.com/stellar/stellar-demo-wallet/blob/master/src/components/wallet/methods/depositAsset.ts) and [withdraw](https://github.com/stellar/stellar-demo-wallet/blob/master/src/components/wallet/methods/withdrawAsset.ts) functionality should be instrumented using the new logview component.   We can use the loggers functionality to add logs for any network calls, instructions, or errors needed to illustrate to a client exactly what's happening during a deposit or withdrawal.

Logger methods can be found [here](https://github.com/stellar/stellar-demo-wallet/blob/master/src/components/logview/logview.tsx#L28-L47) and can be accessed from any method in the wallet component via `this.logger`.  Logs will show up in the log view component that can be opened in the UI with the "Show Logs" button.

msfeldstein:
You can also take a look at sep24.stellar.org as a reference for what we've annotated there.  Stellar-demo-wallet will eventually replace sep24-demo-client so we want to make sure this covers the same things as the demo client.

PR #22, opened 2020-07-15
Introductory PR with a majority of the deposit logging added

lydiat:
### What it is
Initial commit of deposit logging as defined by https://github.com/stellar/stellar-demo-wallet/issues/15

### What got changed
- `depositAsset.ts` - lots of instructions, requests, and responses; I had difficulty including them in the `await` functionality so currently several of the request and response logging is "fake" i.e. it's not tied to actual promises
- `trustAsset.ts` - in order to add `trustline` logging
- `updateAccount.ts` - this may not be necessary as the `updateaccount` function seems to be run multiple times

### What to look for

- Typescript seemingly doesn't like objects in the second argument of the request and response logs, but logview.txt seems like it should allow both types
- Let me know if the "fake" logging is okay or if it should be tied to actual requests
- Currently some of the object parsing in the window is ridiculously long - I'm assuming the clever hide/show functionality for JSON objects that was in the old wallet needs to be transferred?

### DIFF
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -80,14 +80,12 @@ export default async function depositAsset(
     this.logger.instruction(
       'Start the SEP-0010 flow to authenticate the wallet’s Stellar account'
     )
-    const params = { account: this.account.publicKey }
-    this.logger.request(this.toml.WEB_AUTH_ENDPOINT, params)
+    const authParams = { account: this.account.publicKey }
+    this.logger.request(this.toml.WEB_AUTH_ENDPOINT, authParams)
 
     const auth = await axios
       .get(`${toml.WEB_AUTH_ENDPOINT}`, {
-        params: {
-          account: this.account.publicKey,
-        },
+        params: authParams,
       })
       .then(async ({ data }) => {
         this.logger.response(this.toml.WEB_AUTH_ENDPOINT, data)
@@ -147,8 +145,8 @@ export default async function depositAsset(
       'To get the url for the interactive flow check the /transactions/deposit/interactive endpoint'
     )
     this.logger.request(
-      toml.TRANSFER_SERVER_SEP0024 + '/transactions/deposit/interactive',
-      'TODO: form data'
+      `${toml.TRANSFER_SERVER_SEP0024} /transactions/deposit/interactive`
+      /* TODO add form data object */
     )
 
     const interactive = await axios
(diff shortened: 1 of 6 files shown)


msfeldstein:
> Typescript seemingly doesn't like objects in the second argument of the request and response logs, but logview.txt seems like it should allow both types

You can loosen up the types in logview.tsx, for example 

```
interface LogData {
  type: LogDataType
  url?: string
  title?: string
  body?: string | object
}
```
to loosen up the body to accept an object and

```
export interface ILogger {
  request: (url: string, body?: string | object) => Promise<void> | void
  response: (url: string, body?: string | object) => Promise<void> | void
  instruction: (title: string, body?: string) => Promise<void> | void
  error: (title: string, body?: string | object) => Promise<void> | void
}
```
To loosen up the interface and

```
 // Log an outgoing network request with a url and optional body
  @Method()
  async request(url: string, body?: string | object) {
 
```
to loosen up the implementations.  That should let you use the full string or object features of the logview

src/components/wallet/methods/depositAsset.ts
```diff
@@ -57,11 +57,31 @@ export default async function depositAsset(
...
     this.logger.response(infoURL, info)
-    console.log(info)
+
+    this.logger.instruction(
+      'Deposit is enabled, and requires authentication so we should go through SEP-0010'
+    )
+    this.logger.instruction(
+      'Start the SEP-0010 flow to authenticate the wallet’s Stellar account'
+    )
+    const params = { account: this.account.publicKey }
```
msfeldstein, line 83: im a little confused here as these params dont seem to be used anymore, but they definitely should be.  It looks like our sep10 request is missing the account param im actually surprised the existing code works at all

msfeldstein:
> Currently some of the object parsing in the window is ridiculously long - I'm assuming the clever hide/show functionality for JSON objects that was in the old wallet needs to be transferred?

Can you file a bug to add this functionality to the logview?  I wouldn't worry about it for this PR

src/components/wallet/methods/depositAsset.ts
lydiat: They're used on the next line in `this.logger.request(this.toml.WEB_AUTH_ENDPOINT, params)` - is that what you meant? And the `this.account.publicKey` is used in the next network call to the `WEB_AUTH_ENDPOINT` (let me know if I'm misunderstanding the comment!)

msfeldstein: Oh no i just missed that github folded the two relevent lines down so i missed them.  This makes sense.  It would be nice to use the params object you define here inside the params of the axios call just so we ensure that params always stay in sync and reduce duplication

PR #25, opened 2020-07-15
Clean up deposit code

msfeldstein:
I do a few things to make the deposit code easier to read

1) Removed some unnecessary code as commented in PR
2) Removed the use of third party axios as fetch works just fine
3) Reduced the usage of `.then()` promise chaining in favor of using `await` for more linear code progression
4) Added some error case handling, the last PR added all the happy path logging but we should bail early with an error when something goes wrong

More details in the comments

### DIFF
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -1,11 +1,6 @@
 import { Transaction, Server } from 'stellar-sdk'
-import axios from 'axios'
-import {
-  get as loGet,
-  each as loEach,
-  findIndex as loFindIndex,
-} from 'lodash-es'
 import TOML from 'toml'
+import { get } from 'lodash-es'
 
 import { handleError } from '@services/error'
 import { stretchPincode } from '@services/argon2'
@@ -20,6 +15,7 @@ export default async function depositAsset(
   const loadingKey = `deposit:${asset_code}:${asset_issuer}`
   this.loading = { ...this.loading, [loadingKey]: true }
   const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
+
   try {
     const pincode = await this.setPrompt({
       message: 'Enter your account pincode',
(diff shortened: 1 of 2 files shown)


src/components/wallet/methods/depositAsset.ts
```diff
@@ -30,25 +26,14 @@ export default async function depositAsset(
...
-      asset_issuer,
-    })
+    this.logger.request('Fetch issuer account from Horizon', asset_issuer)
+    const issuerAccount = await this.server.loadAccount(asset_issuer)
+    this.logger.response('Fetch issuer account from Horizon', issuerAccount)
+    const homeDomain = issuerAccount.home_domain
+    if (!homeDomain) throw "Couldn't find a home_domain on the assets issuer"
 
-    if (hasCurrency === -1)
-      await this.trustAsset(asset_code, asset_issuer, pincode_stretched)
```
msfeldstein, line 40: Now that we use your existing balances to start the deposit flow, there's no way to get here without already having a trustline in place.

src/components/wallet/methods/depositAsset.ts
```diff
@@ -30,25 +26,14 @@ export default async function depositAsset(
...
-
-    const server = this.server as Server
-    const issuerAccount = await server.loadAccount(asset_issuer)
-    const homeDomain: string = (issuerAccount as any).home_domain
```
msfeldstein, line 44: We can remove a good deal of the Typescript type casting as we now have proper types in `this`, due to the 'fake this' feature of typescript.  We pass a value called `this` to the function, and typescript will use that type to infer the true type of `this`.

src/components/wallet/methods/depositAsset.ts
```diff
@@ -30,25 +26,14 @@ export default async function depositAsset(
...
-    const balances = loGet(this.account, 'state.balances')
-    const hasCurrency = loFindIndex(balances, {
-      asset_code,
-      asset_issuer,
-    })
+    this.logger.request('Fetch issuer account from Horizon', asset_issuer)
+    const issuerAccount = await this.server.loadAccount(asset_issuer)
+    this.logger.response('Fetch issuer account from Horizon', issuerAccount)
+    const homeDomain = issuerAccount.home_domain
+    if (!homeDomain) throw "Couldn't find a home_domain on the assets issuer"
```
msfeldstein, line 33: All errors can be handled by the `catch` at the bottom

src/components/wallet/methods/depositAsset.ts
```diff
@@ -61,147 +46,129 @@ export default async function depositAsset(
       `Received WEB_AUTH_ENDPOINT from TOML: ${toml.WEB_AUTH_ENDPOINT}`
     )
     this.logger.instruction(
-      `Received TRANSFER_SERVER from TOML: ${toml.TRANSFER_SERVER}`
+      `Received TRANSFER_SERVER_SEP0024 from TOML: ${toml.TRANSFER_SERVER_SEP0024}`
```
msfeldstein, line 49: We updated SEP24 at some point to use the key TRANSFER_SERVER_SEP0024 instead of TRANSFER_SERVER so that it doesn't conflict with SEP6's TRANSFER_SERVER key

src/components/wallet/methods/depositAsset.ts
```diff
@@ -213,7 +180,7 @@ export default async function depositAsset(
       }
     }
   } catch (err) {
-    this.loading = { ...this.loading, deposit: false }
+    finish()
```
msfeldstein, line 183: Always use the finish function so we don't run into problems where we've set the wrong loading key to false, and end up with a non-stop spinner.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -29,7 +29,7 @@ export default async function updateAccount(this: Wallet) {
             'paging_token',
           ]),
         }
-        this.logger.response(selfURL.self.href, account)
+        // this.logger.response(selfURL.self.href, account)
```
msfeldstein, line 32: I want to comment these out until we get the foldable JSON just so we don't start off with a lot of noise

src/components/wallet/methods/depositAsset.ts
JakeUrban: I'm not sure I understand this. What do you mean by existing balances? Can I not provide my own account for the demo wallet to use?

msfeldstein: yeah you can, i just mean you can't start a deposit for a token you don't already have a trustline for, because there won't be a balance row for it which has the buttons in it.

This was originally for when the wallet had hard-coded the SRT token

PR #26, opened 2020-07-16
Add collapsible JSON in logview

msfeldstein:
The logviews were huge and noisy since we just serialized the full json objects.  This PR adds a json-viewer webcomponent to allow us to show only the top level of an object and allow you to expand subproperties to view more detail.

The code in components.d.ts is auto-generated you don' t need to review that.

Screenshot 2020-07-16 at 11 25 33 AM

### DIFF
--- a/src/components/jsonviewer/jsonviewer.tsx
+++ b/src/components/jsonviewer/jsonviewer.tsx
@@ -1,0 +1,43 @@
+import { Component, h, Prop } from '@stencil/core'
+
+@Component({
+  tag: 'json-viewer',
+  styleUrl: 'jsonviewer.scss',
+  shadow: true,
+})
+export class JSONViewer {
+  @Prop() data: any
+
+  render() {
+    return (
+      <div>
+        {'{'}
+        <div class="indent">
+          {Object.keys(this.data).map((key) => {
+            const val = this.data[key]
+            let display = ''
+            switch (typeof val) {
+              case 'object':
+                display = (
+                  <details>
+                    <summary>
+                      {key}: {Object.keys(val).length} values
+                    </summary>
+                    <json-viewer data={val}></json-viewer>
+                  </details>
+                )
+                break
+              case 'function':
+                display = `${key}: Function`
+                break
+              default:
+                display = `${key}: ${val}`
+            }
+            return <div>{display}</div>
+          })}
+        </div>
+        {'}'}
+      </div>
+    )
+  }
+}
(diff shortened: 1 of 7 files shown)


src/components/jsonviewer/jsonviewer.tsx
```diff
@@ -0,0 +1,43 @@
+import { Component, h, Prop } from '@stencil/core'
+
+@Component({
+  tag: 'json-viewer',
+  styleUrl: 'jsonviewer.scss',
+  shadow: true,
```
msfeldstein, line 6: The shadow property means that this web component is fully isolated from the rest of the doucments styling.  Nothing i put in this files css file affects the rest of the page, and this element is not affected by any other css in the page.

src/components/jsonviewer/jsonviewer.tsx
```diff
@@ -0,0 +1,43 @@
...
+})
+export class JSONViewer {
+  @Prop() data: any
+
+  render() {
+    return (
+      <div>
+        {'{'}
```
msfeldstein, line 14: I wish there was a nicer way to put a curly brace in my component, but using a js string of it is the best bet right now.

src/components/logview/logview.tsx
```diff
@@ -8,7 +8,6 @@ enum LogDataType {
 }
 interface LogData {
   type: LogDataType
-  url?: string
```
msfeldstein, line 11: I wanted to standarize the LogData type, there's no reason to differentiate between `url` and `title` property, they both act as titles.

src/components/logview/logview.tsx
```diff
@@ -65,50 +64,24 @@ export class LogView {
...
-
-      default:
-        return null
-    }
+    const className = {
+      [LogDataType.Instruction]: 'instruction',
+      [LogDataType.Error]: 'error',
+      [LogDataType.Request]: 'request',
+      [LogDataType.Response]: 'response',
+    }[data.type]
```
JakeUrban, line 72: Won't this be a KeyError since the keys are wrapped in lists?

msfeldstein: This syntax is a way of using variables as the keys for objects

```
const a = "Hello";
const obj1 = {
  a: "World"
}
const obj2 = {
  [a]: "World"
}
```
Obj1 looks like `{a: "World"}` while obj2 looks like `{Hello: "World"}`

PR #29, opened 2020-07-16
Clean up the trust asset flow

msfeldstein:
Use explicit variable names for asset_code and issuer, vs instructions[0][1]
Break out nested promises into async
Add proper logger calls

### DIFF
--- a/src/components/wallet/methods/trustAsset.ts
+++ b/src/components/wallet/methods/trustAsset.ts
@@ -18,24 +18,14 @@ export default async function trustAsset(
   issuer?: string,
   pincode_stretched?: Uint8Array
 ) {
+  this.loading = { ...this.loading, trust: true }
+  this.error = null
+  console.log(this.loading)
+  const finish = () => (this.loading = { ...this.loading, trust: false })
   try {
-    let instructions
-
-    this.logger.instruction(
-      'We need to add a trustline to the asset to ensure the deposit will be expected'
-    )
-
-    if (asset && issuer) {
-      instructions = [asset, issuer]
-      this.logger.instruction(
-        'There is already a trustline on this account, no need to recreate it'
-      )
-    } else {
-      this.logger.instruction(
-        'There isn’t currently a trustline on this account so we need to add one'
-      )
-      instructions = await this.setPrompt({ message: '{Asset} {Issuer}' })
-      instructions = instructions.split(' ')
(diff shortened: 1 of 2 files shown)


src/components/wallet/methods/trustAsset.ts
```diff
@@ -18,24 +18,14 @@ export default async function trustAsset(
...
-      )
-    } else {
-      this.logger.instruction(
-        'There isn’t currently a trustline on this account so we need to add one'
-      )
-      instructions = await this.setPrompt({ message: '{Asset} {Issuer}' })
-      instructions = instructions.split(' ')
+    if (!asset || !issuer) {
+      let instructions = await this.setPrompt({ message: '{Asset} {Issuer}' })
+      ;[asset, issuer] = instructions.split(' ')
```
JakeUrban, line 28: wrong line for the semicolon

src/components/wallet/methods/trustAsset.ts
```diff
@@ -52,37 +42,33 @@ export default async function trustAsset(
...
-      .then((res) => console.log(res))
-      .finally(() => {
-        this.loading = { ...this.loading, trust: false }
-        this.updateAccount()
-      })
+    transaction.sign(keypair)
+    this.logger.request('Submitting changeTrust transaction', transaction)
+    const result = await this.server.submitTransaction(transaction)
+    this.logger.response('Submitted changeTrust transaction', result)
+    await this.updateAccount()
```
JakeUrban, line 66: What does updateAccount do? update account balances we have in-memory?

JakeUrban:
By the way, I don't see any tests for this app. We should add the skeleton/tools needed for that soon.

src/components/wallet/methods/trustAsset.ts
msfeldstein: for some reason the auto-formatter really wants it there... we don't usually use semicolons but in this case, where we do a spread and the line starts with an open brackets, its needed. I don't know why the formatter keeps moving it down there

src/components/wallet/methods/trustAsset.ts
msfeldstein: yeah it updates balances from horizon, rather than try to make discrete updates based on the action taken

msfeldstein:
agreed on tests, can look into that soon

PR #36, opened 2020-11-03
Remove Pincode

JakeUrban:
Remove the step where the wallet asks for your pincode.

This was necessary originally because the application stored an encrypted hash of the user's secret key. The secret key has been added to the `Wallet` object and is now stored unencrypted in the browser's local storage.


Zagan202 (approved):
LGTM
Works as intended!

PR #37, opened 2020-11-04
Add multiple inputs to Prompt component

JakeUrban:
Doesn't change any functionality from the user's perspective. UX is unchanged.

The difference under the hood is that we can now ask for multi-input prompts. We'll use this at first to ask for the asset code, issuer, or homedomain in separate input fields.

### DIFF
--- a/src/components/wallet/methods/trustAsset.ts
+++ b/src/components/wallet/methods/trustAsset.ts
@@ -19,10 +19,11 @@ export default async function trustAsset(
   this.error = null
   console.log(this.loading)
   const finish = () => (this.loading = { ...this.loading, trust: false })
+  let inputs
   try {
     if (!asset || !issuer) {
-      let instructions = await this.setPrompt({ message: '{Asset} {Issuer}' })
-      ;[asset, issuer] = instructions.split(' ')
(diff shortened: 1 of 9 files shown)


src/components/wallet/methods/trustAsset.ts
```diff
@@ -19,10 +19,11 @@ export default async function trustAsset(
   this.error = null
   console.log(this.loading)
   const finish = () => (this.loading = { ...this.loading, trust: false })
+  let inputs
   try {
     if (!asset || !issuer) {
-      let instructions = await this.setPrompt({ message: '{Asset} {Issuer}' })
-      ;[asset, issuer] = instructions.split(' ')
+      inputs = await this.setPrompt({ message: '{Asset} {Issuer}' })
+      ;[asset, issuer] = inputs[0].value.split(' ')
```
Zagan202, line 26: Is there a way to leverage the new multi-input prompt here instead of adopting the split(' ') logic from before?

Zagan202 (approved):
LGTM

I do have one question but I do see the potential of this change!

src/components/wallet/methods/trustAsset.ts
JakeUrban: Yes, thats my next step PR haha

PR #38, opened 2020-11-04
Fix tomlURL bug

Zagan202:
- there was a case where the URL builder would fail if there wasnt a
`https://` attached to it

Signed-off-by: Lawrence Lawson <lawrence@stellar.org>

### DIFF
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -24,7 +24,6 @@ export default async function depositAsset(
     this.logger.instruction(
       `Found home_domain '${homeDomain}' as issuer's domain, fetching the TOML file to find the transfer server`
     )
-    debugger
     const tomlURL = new URL(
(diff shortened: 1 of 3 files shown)


src/components/wallet/methods/depositAsset.ts
```diff
@@ -24,7 +24,10 @@ export default async function depositAsset(
     this.logger.instruction(
       `Found home_domain '${homeDomain}' as issuer's domain, fetching the TOML file to find the transfer server`
     )
-    const tomlURL = new URL(homeDomain)
+    debugger
```
JakeUrban, line 27:
```suggestion
```

JakeUrban (approved):

src/components/wallet/methods/depositAsset.ts
Zagan202: whoops

Issue #34, opened 2020-10-23
Implement the claimable balance support in our demo client/wallet


Zagan202:
### What problem does your feature solve?

In order to leverage the new Protocol 14/15 feature claimable balances we need wallets to query for accounts that have claimable balances to claim.

### What would you like to see?

Notify Wallet users/accounts when a claimable balance is available for them to claim.

Zagan202:
We should also deprecate the original demo client and only add new features to the new one.
This may need to be separated into another issue(s) entirely.

tyvdh:
What is "the original demo client"?

Also another aside is does it make sense to keep this repo for anything other than the tutorial now that the new account viewer is nearly ready? Should we be using that for all future demo/reference stuff? This may be a bigger conversation but I think it's an important one to have before we dupe too much between this repo and the new AV.

cc: @morleyzhi

Zagan202:
Hey @tyvdh, I need to touch base with @JakeUrban when he gets back to specify which client "demo client" he's [referring to](https://github.com/stellar/stellar-protocol/issues/747#issuecomment-712447987) when he comes back on monday.

I believe we're referring to https://github.com/stellar/sep24-demo-client 

> Should we be using that for all future demo/reference stuff? 

The goal we would like to achieve is use the stellar-demo-wallet and polaris as an end to end experience demo and reduce duplication. If we want to extend the or edit either SEP we dont want to change both stellar-demo-wallet and sep24-demo-client.

Zagan202:
> This may be a bigger conversation but I think it's an important one to have before we dupe too much between this repo and the new AV.

Im open to have this discussion with all relevant contributors sometime next week.
I agree we should be all on the same page for any big changes.

JakeUrban:
Hi @tyvdh, can you explain your reasoning here? Account Viewer is not really a mock wallet that supports our SEP protocols, its an exploratory tool for looking at accounts on Stellar, so in my team's view they're completely disjoint areas.

@Zagan202 you're right, https://github.com/stellar/sep24-demo-client is the currently-used demo wallet but we need to transition to using this project for that purpose. 

Most of the ecosystem knows of the old demo wallet, so we need to ensure it's functional until they've transitioned to the new one. That transition process still needs to be worked out. I'm a fan of simply replacing the old client with the new one on the same domain once we're confident the new one is ready, whatever that means.

tyvdh:
[...]

PR #39, opened 2020-11-05
Claimable balance Query feature

Zagan202:
# Description

Extends the Update Account feature to include querying claimable balances for the current account.
`updateAccount` is periodically called whenever the wallet triggers one of the following components
- createAccount 
- loadAccount
- updateAccount 
- depositAsset
- withdrawAsset 
- trustAsset
- makePayment
- makeRegulatedPayment

~Currently displays each available claimable balance in the logging panel on the right of the UI~
Currently displays each available claimable balance when you click the `Show Claimable Balance Details` button.

## TODO?

Maybe adding support to claim the balances from within the wallet is something we want in include.

Addresses issue: #34 

## Type of change

# How Has This Been Tested?

1. `npm i`
2. `npm start`
3. If you want to see a an account with claimable balances available load this account `SBK52ZEYXQLOKC4Y646NQFJGCVYLXWQYVTA6TU7QKPAUL2F45X6NABM2`

### DIFF
--- a/src/components/wallet/methods/updateAccount.ts
+++ b/src/components/wallet/methods/updateAccount.ts
@@ -7,27 +7,28 @@ import {
 import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
+export function claimableBalanceQuery({ this: wallet }) {
+  wallet.server
+    .claimableBalances()
+    .claimant(wallet.account.publicKey)
+    .call()
+    .then(function (resp) {
+      resp.records.forEach((record) => {
+        wallet.logger.response(
+          'Claimable Balances Available ',
+          loOmit(record, ['_links', 'paging_token', 'self'])
+        )
+      })
+      console.log(resp)
+    })
+    .catch(function (err) {
+      wallet.error = handleError(err)
+      console.error(err)
+    })
+  wallet.loading = { ...wallet.loading, update: false }
+}
+
 export default async function updateAccount(this: Wallet) {
-  function cbQuery({ this: wallet }) {
-    wallet.server
-      .claimableBalances()
-      .claimant(wallet.account.publicKey)
-      .call()
-      .then(function (resp) {
-        resp.records.forEach((record) => {
-          wallet.logger.response(
-            'Claimable Balances Available ',
-            loOmit(record, ['_links', 'paging_token', 'self'])
-          )
-        })
-        console.log(resp)
-      })
-      .catch(function (err) {
-        wallet.error = handleError(err)
-        console.error(err)
-      })
-    wallet.loading = { ...wallet.loading, update: false }
-  }
   try {
     this.error = null
     this.loading = { ...this.loading, update: true }
@@ -46,7 +47,7 @@ export default async function updateAccount(this: Wallet) {
           ]),
         }
       })
-      .finally(() => cbQuery({ this: this }))
+      .finally(() => claimableBalanceQuery({ this: this }))
   } catch (err) {
     this.error = handleError(err)
   }
--- a/src/components/wallet/methods/updateAccount.ts
+++ b/src/components/wallet/methods/updateAccount.ts
@@ -8,39 +8,34 @@ import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
 export default async function updateAccount(this: Wallet) {
-  function cbQuery({this: wallet}) {
+  function cbQuery({ this: wallet }) {
     wallet.server
       .claimableBalances()
       .claimant(wallet.account.publicKey)
       .call()
       .then(function (resp) {
-        resp.records.forEach(record => {
-          wallet.logger.response("Claimable Balances Available ",loOmit(record, [
-            '_links',
-            'paging_token',
-            'self'
-          ]))
-        });
-        console.log(resp);
+        resp.records.forEach((record) => {
+          wallet.logger.response(
+            'Claimable Balances Available ',
+            loOmit(record, ['_links', 'paging_token', 'self'])
+          )
+        })
+        console.log(resp)
       })
       .catch(function (err) {
         wallet.error = handleError(err)
-        console.error(err);
-      });
+        console.error(err)
+      })
     wallet.loading = { ...wallet.loading, update: false }
-  };
+  }
   try {
     this.error = null
     this.loading = { ...this.loading, update: true }
-    // this.logger.instruction('Updating account...')
     await this.server
       .accounts()
       .accountId(this.account.publicKey)
       .call()
       .then((account) => {
-        // const selfURL = loFind(account, 'self')
-        // this.logger.request(selfURL.self.href)
-
         this.account = {
           ...this.account,
           state: loOmit(account, [
@@ -50,9 +45,8 @@ export default async function updateAccount(this: Wallet) {
             'paging_token',
           ]),
         }
-        // this.logger.response(selfURL.self.href, account)
       })
-      .finally(() => (cbQuery({this:this})))
+      .finally(() => cbQuery({ this: this }))
   } catch (err) {
     this.error = handleError(err)
   }
--- a/src/components/wallet/methods/updateAccount.ts
+++ b/src/components/wallet/methods/updateAccount.ts
@@ -7,25 +7,21 @@ import {
 import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
-export function claimableBalanceQuery({ this: wallet }) {
-  wallet.server
-    .claimableBalances()
-    .claimant(wallet.account.publicKey)
-    .call()
-    .then(function (resp) {
-      resp.records.forEach((record) => {
-        wallet.logger.response(
-          'Claimable Balances Available ',
-          loOmit(record, ['_links', 'paging_token', 'self'])
-        )
-      })
-      console.log(resp)
+async function claimableBalanceQuery(wallet: Wallet) {
+  try {
+    let resp = await wallet.server
+      .claimableBalances()
+      .claimant(wallet.account.publicKey)
+      .call()
+    resp.records.forEach((record) => {
+      wallet.logger.response(
+        'Claimable Balances Available ',
+        loOmit(record, ['_links', 'paging_token', 'self'])
+      )
     })
-    .catch(function (err) {
-      wallet.error = handleError(err)
-      console.error(err)
-    })
-  wallet.loading = { ...wallet.loading, update: false }
+  } catch (err) {
+    wallet.error = handleError(err)
+  }
 }
 
 export default async function updateAccount(this: Wallet) {
@@ -47,7 +43,11 @@ export default async function updateAccount(this: Wallet) {
           ]),
         }
       })
-      .finally(() => claimableBalanceQuery({ this: this }))
+      .finally(() =>
+        claimableBalanceQuery(this).finally(() => {
+          this.loading = { ...this.loading, update: false }
+        })
+      )
   } catch (err) {
     this.error = handleError(err)
   }
--- a/src/components/wallet/methods/updateAccount.ts
+++ b/src/components/wallet/methods/updateAccount.ts
@@ -19,12 +19,12 @@ export default async function updateAccount(this: Wallet) {
       .claimableBalances()
       .claimant(this.account.publicKey)
       .call()
-    // claimableBalanceResp.records.forEach((record) => {
-    //   this.logger.response(
-    //     'Claimable Balances Available ',
-    //     loOmit(record, ['_links', 'paging_token', 'self'])
-    //   )
-    // })
+    claimableBalanceResp.records.forEach((record) => {
+      this.logger.response(
+        'Claimable Balances Available ',
+        loOmit(record, ['_links', 'paging_token', 'self'])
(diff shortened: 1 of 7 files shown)


stencil.config.ts
```diff
@@ -30,6 +30,7 @@ export const config: Config = {
   commonjs: {
     namedExports: {
       'stellar-sdk': [
+        'Server',
```
JakeUrban, line 33: What is this stuff? Its looks like we have to hard-code the pieces of our stellar-sdk we use, which feels weird/wrong. If `Server` is not included, can I access `Server` in our JS code? I guess I don't know what this config file is.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
 export default async function updateAccount(this: Wallet) {
+  function cbQuery({ this: wallet }) {
+    wallet.server
+      .claimableBalances()
+      .claimant(wallet.account.publicKey)
+      .call()
+      .then(function (resp) {
```
JakeUrban, line 16:
```suggestion
      .then((resp) => {
```

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
...
+        resp.records.forEach((record) => {
+          wallet.logger.response(
+            'Claimable Balances Available ',
+            loOmit(record, ['_links', 'paging_token', 'self'])
+          )
+        })
+        console.log(resp)
```
JakeUrban, line 23: We shouldn't do this right? I assume this was for your debugging purposes?

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
...
+      })
+      .catch(function (err) {
+        wallet.error = handleError(err)
+        console.error(err)
```
JakeUrban, line 27: I would image `handleError()` pushes the message to the relevant channels right? `console.error()` probably isn't needed.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
...
+      })
+    wallet.loading = { ...wallet.loading, update: false }
```
JakeUrban, line 29: I'm not a fan of setting `update: false` here when `update: true` is set in the outer function. I made a comment where `cbQuery` is called about how to change this.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
 export default async function updateAccount(this: Wallet) {
+  function cbQuery({ this: wallet }) {
+    wallet.server
```
JakeUrban, line 12: So, I see you're using the old style of javascript of chaining callbacks. Very understandable, I did that too when I first started working on Michael's JS projects.

What I learned is that ES6 added a bunch of native async capabilities that make these much more readable.

For example, your current `wallet.server.call()` code can be converted to this:
```
let resp = await wallet.server.claimableBalances().claimant(wallet.account.publicKey).call()
```
You can then use `resp` on the next line like so:
```
resp.records.forEach((record) => {
    wallet.logger.response(...)
})
```
Essentially, you can get rid of the deep nesting of javascript code within other function calls. It makes JS code so much more readable, so please use this style from hereon out!

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
...
+      .then(function (resp) {
+        resp.records.forEach((record) => {
+          wallet.logger.response(
+            'Claimable Balances Available ',
+            loOmit(record, ['_links', 'paging_token', 'self'])
+          )
+        })
+        console.log(resp)
+      })
+      .catch(function (err) {
```
JakeUrban, line 25:
```suggestion
      .catch((err) => {
```

JakeUrban (changes requested):
Good stuff, the functionality looks good. The code organization and JS syntax need some adjustments.

Also as a general, there's no need to keep function names overly-short. `cb` is a short hand for claimable balances that we've kinda created on the fly. Can we change the function name from `cbQuery` to `claimableBalanceQuery`?

stencil.config.ts
Zagan202: Long story short Stencil needs you to explicitly mark (_some?_) what classes you're exporting from modules you've imported.
Something to do with rollout but I'm not an expert.

I do know that it is what's making things function currently.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -8,18 +8,34 @@ import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
 
 export default async function updateAccount(this: Wallet) {
+  function cbQuery({ this: wallet }) {
```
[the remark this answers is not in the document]
Zagan202: Great points implemented now!

src/components/wallet/methods/updateAccount.ts
Zagan202: yeah this will be removed

src/components/wallet/methods/updateAccount.ts
Zagan202: True your suggestion make much more sense

src/components/wallet/methods/updateAccount.ts
Zagan202: Check out the refactor here https://github.com/stellar/stellar-demo-wallet/pull/39/commits/fabd6cc73f8e791ca2d2eb64a8692c7951363c20

src/components/wallet/methods/updateAccount.ts
```diff
@@ -7,19 +7,32 @@ import {
...
 export default async function updateAccount(this: Wallet) {
   try {
     this.error = null
     this.loading = { ...this.loading, update: true }
-    // this.logger.instruction('Updating account...')
     await this.server
       .accounts()
       .accountId(this.account.publicKey)
       .call()
       .then((account) => {
```
JakeUrban, line 35: Can we adjust this call too? I'm a bit surprised this even works. If you use `await`, that means you're expecting the code that follows to return a JS `Promise`. But this call ends with a `then()` clause that I don't think returns a `Promise` since its not marked as async like so:
```
.then(async (account) => {
  ...   
}
```

JakeUrban: So maybe this actually works, idk, but it looks weird and it doesn't match the pattern you now follow in `claimableBalanceQuery()`

src/components/wallet/methods/updateAccount.ts
```diff
@@ -7,19 +7,32 @@ import {
...
+      .claimableBalances()
+      .claimant(wallet.account.publicKey)
+      .call()
+    resp.records.forEach((record) => {
+      wallet.logger.response(
+        'Claimable Balances Available ',
+        loOmit(record, ['_links', 'paging_token', 'self'])
+      )
+    })
+  } catch (err) {
```
JakeUrban, line 22: I think we can make this function simpler by removing the try/catch and letting `updateAccount` call `handleError()`

src/components/wallet/methods/updateAccount.ts
JakeUrban: Take a look at this site: https://javascript.info/async-await

See how they replace ALL callbacks (`.finally()`, `.then()`, and `.catch()`)? I'd like use to move to this style so we can avoid thinking about what kind of error catching we need (`try {} catch {}` vs `.then() and .catch()`

JakeUrban: Its really just a lot easier to read and understand JS when it doesn't have crazy callback chains

stencil.config.ts
JakeUrban: I almost missed your explanation! Let the person asking the question figure out if its resolved.

And ok sounds good, don't fix whats not broken. But this does seem a little sus

JakeUrban (changes requested):
90% of the way there! Just want to make sure we're getting our first major feature right, not just in terms of functionality but in terms of style and dev best practices.

JakeUrban:
I also just noticed I'm not sure what the end result is of this change. It looks like claimable balances are just logged on the righthand side. Could we add the claimable balances JSON objects to the Account Details JSON displayed when you press the `Show Account Details` button?

src/components/wallet/methods/updateAccount.ts
Zagan202: I think the changes I made align more with your change request.
It does make the code more clean!

src/components/wallet/methods/updateAccount.ts
Zagan202: Good call

Zagan202:
> I also just noticed I'm not sure what the end result is of this change. It looks like claimable balances are just logged on the righthand side. Could we add the claimable balances JSON objects to the Account Details JSON displayed when you press the `Show Account Details` button?

I think that is entirely possible.
Surfacing that information in the logger was more of a POC intermediate change.

JakeUrban (approved):
Beautiful. The one thing I'd say is that we should follow this PR up with actually getting these claimable balances in Account Details view.

Zagan202:
Currently displays each available claimable balance when you click the `Show Claimable Balance Details` button.

src/components/wallet/methods/updateAccount.ts
```diff
@@ -28,11 +15,21 @@ export default async function updateAccount(this: Wallet) {
       .accounts()
       .accountId(this.account.publicKey)
       .call()
+    let claimableBalanceResp = await this.server
+      .claimableBalances()
+      .claimant(this.account.publicKey)
+      .call()
+    // claimableBalanceResp.records.forEach((record) => {
```
JakeUrban, line 22: Hmm so I think claimable balance should still be logged so the user knows they have claimable balances outstanding. If they aren't logged, there is no visual indication that they exist right? I guess ultimately we'll have a button for claiming them, which should have a visual indicator that the claimable balance is available.

src/components/wallet/wallet.ts
```diff
@@ -40,6 +40,7 @@ interface StellarAccount {
   publicKey: string
   secretKey: string
   state?: ServerApi.AccountRecord
+  claimableBalances?: ServerApi.ClaimableBalanceRecord
```
JakeUrban, line 43: What is `ServerApi`?, it's not great that things are usable in a file without being imported.

src/components/wallet/views/loggedInContent.tsx
```diff
@@ -55,5 +55,17 @@ export default function loggedInContent(this: Wallet) {
         </pre>
       </collapsible-container>
     ) : null,
+
+    loHas(this.account, 'state') ? (
+      <collapsible-container
+        id="account-details-container"
```
JakeUrban, line 61: Are you sure we can have 2 `collapsible-container` components with the same id?

src/components/wallet/views/loggedInContent.tsx
```diff
@@ -55,5 +55,17 @@ export default function loggedInContent(this: Wallet) {
         </pre>
       </collapsible-container>
     ) : null,
+
+    loHas(this.account, 'state') ? (
```
JakeUrban, line 59: Can you explain what `loHas` and `loOmit` are? 

Based on what I'm understanding, I think this line needs to be changed to:
```suggestion
    loHas(this.account, 'claimableBalances') ? (
```

Otherwise the component will be rendered whenever `state` changes, not `claimableBalances` right?

src/components/wallet/methods/updateAccount.ts
Zagan202: Yes I'm happy to display the balances in the logger as well.
I thought at first it was a bit redundant since you can see the same results with the `Show Claimable Balance Details` button.

> I guess ultimately we'll have a button for claiming them, which should have a visual indicator that the claimable balance is available.

I do think that there is a path where we dont have this  `Show Claimable Balance Details` button and instead just have `"Claim Outstanding Balances"` that will show and allow the user to claim them.

src/components/wallet/wallet.ts
Zagan202: I was more or less following the logic that Tyler set up for this interface.
The `ServerApi` is imported on line 2 of the file.

src/components/wallet/views/loggedInContent.tsx
Zagan202: Apologies still getting used css things.

src/components/wallet/views/loggedInContent.tsx
Zagan202: Good catch!

src/components/wallet/wallet.ts
JakeUrban: Oh I missed that, whoops

JakeUrban (approved):

PR #44, opened 2020-11-06
Add claimable_balance_supported to postDepositParams

Zagan202:


JakeUrban (approved):
are we going with `is_claimable_...` or just `claimable_...`?

Zagan202:
> are we going with `is_claimable_...` or just `claimable_...`?

 `claimable_...`

PR #45, opened 2020-11-06
Add ability to add asset by home domain

JakeUrban:
Screen Shot 2020-11-06 at 2 30 10 PM


Zagan202 (changes requested):
LGTM but one request.

Currently, on master and on your branch, when you cancel a `trustAsset` call (hitting the cancel button) it gives an error in the logger.

It would be nice if it handled canceling a bit more gracefully.

JakeUrban:
So unfortunately every call to `setPrompt()` will currently have this issue if `cancel` is selected. `trustAsset` will be the first operation to cancel gracefully.

Zagan202:
> So unfortunately every call to `setPrompt()` will have this issue if `cancel` is selected. `trustAsset` will be the first operation to cancel gracefully. Should I take the extra X minutes to do this with every event?

Let's create a separate issue for that to not block this PR.

JakeUrban:
https://github.com/stellar/stellar-demo-wallet/issues/47

Zagan202 (commented):
Official LGTM

Issue #42, opened 2020-11-06
Feature Request: Claim Balances Button


Zagan202:
### What problem does your feature solve?

Give users the ability to claim any outstanding claimable balances.

### What would you like to see?

- Ability to claim claimable balances  
- Easy to follow UX 
  - Surface what balances could be claimed
  - Display what conditions (predicates) are required to claim said balances if any
  - Sign to confirm claim.

- Deprecate `Show Claimable Balance Details` in favor of `Claim Available Balances`
  - Nice to have: Show `Claim Available Balances` only when a user has balances available

PR #49, opened 2020-11-11
Claim ClaimableBalance Button

Zagan202:
# Description

Adds a new Claim button and displays Claimable balances in their own view.

- Show claimable balance `{asset_code: amount}` Below the account's current asset balances
- Add Claim button to add balance to account's current asset balances
- Remove `Show Claimable Balances` button
    - `Claimable Balances Available` will still show on the logger (Right Side)

Addresses issue: #42

## Type of change

# How Has This Been Tested?

1. `npm start`
2.  If you want to see a an account with claimable balances available load this account 
    - `SBK52ZEYXQLOKC4Y646NQFJGCVYLXWQYVTA6TU7QKPAUL2F45X6NABM2`
    - You should see 3 balances on this account you could claim the `LAWR` balances but the `SRT` should fail due to no trust line established to SRT at the time of writing this.

### DIFF
--- a/src/components/wallet/methods/claimAsset.ts
+++ b/src/components/wallet/methods/claimAsset.ts
@@ -7,23 +7,23 @@ import {
   Keypair,
 } from 'stellar-sdk'
 import { handleError } from '@services/error'
+import { ClaimableBalance } from '../views/claimableDisplay'
 import { Wallet } from '../wallet'
 
 export default async function claimAsset(
   this: Wallet,
-  balanceId: string,
-  assetCode: string,
-  sponsor: string,
-  amount: string
+  balance: ClaimableBalance
 ) {
-  const loadingKey = `claim:${assetCode}:${balanceId}`
+  const balanceId = balance.id
+  const assetCode =
+    balance.asset === 'native' ? 'XLM' : balance.asset.split(':')[0]
+  const loadingKey = `claim:${assetCode}:${balance.id}`
   this.loading = { ...this.loading, [loadingKey]: true }
   const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
-
   try {
     this.logger.instruction(
-      `Claiming ${amount} of ${assetCode}`,
-      `BalanceId: ${balanceId} Sponsor:${sponsor}`
+      `Claiming ${balance.amount} of ${assetCode}`,
+      `BalanceId: ${balance.id} Sponsor:${balance.sponsor}`
     )
     const keypair = Keypair.fromSecret(this.account.secretKey)
     const accountRecord = await this.server
--- a/src/components/wallet/methods/claimAsset.ts
+++ b/src/components/wallet/methods/claimAsset.ts
@@ -1,0 +1,63 @@
+import {
+  Account,
+  TransactionBuilder,
+  BASE_FEE,
+  Networks,
+  Operation,
+  Keypair,
+} from 'stellar-sdk'
+import { handleError } from '@services/error'
+import { Wallet } from '../wallet'
+
+export default async function claimAsset(
+  this: Wallet,
+  balanceId: string,
+  assetCode: string,
+  sponsor: string,
+  amount: string
+) {
+  const loadingKey = `claim:${assetCode}:${balanceId}`
+  this.loading = { ...this.loading, [loadingKey]: true }
+  const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
+
+  try {
+    this.logger.instruction(
+      `Claiming ${amount} of ${assetCode}`,
+      `BalanceId: ${balanceId} Sponsor:${sponsor}`
+    )
+    const keypair = Keypair.fromSecret(this.account.secretKey)
+    const accountRecord = await this.server
+      .accounts()
+      .accountId(keypair.publicKey())
+      .call()
+    this.logger.instruction(
+      'Loading account to get sequence number for claimClaimableBalance transaction'
+    )
+    const account = new Account(keypair.publicKey(), accountRecord.sequence)
+    this.logger.instruction('Building claimClaimableBalance transaction')
+    const transaction = new TransactionBuilder(account, {
+      fee: BASE_FEE,
+      networkPassphrase: Networks.TESTNET,
+    })
+      .addOperation(
+        Operation.claimClaimableBalance({
+          balanceId,
+        })
+      )
+      .setTimeout(0)
+      .build()
+    transaction.sign(keypair)
+    this.logger.request('Submitting changeTrust transaction', transaction)
+    const result = await this.server.submitTransaction(transaction)
+    this.logger.response('Submitted claimClaimableBalance transaction', result)
+    await this.updateAccount()
+    finish()
+  } catch (err) {
+    this.logger.error(
+      'Error in claimClaimableBalance transaction',
+      err.toJSON()
+    )
+    this.error = handleError(err.response.data.extras.result_codes.operations)
+    finish()
+  }
+}
--- a/src/components/wallet/methods/claimAsset.ts
+++ b/src/components/wallet/methods/claimAsset.ts
@@ -47,7 +47,10 @@ export default async function claimAsset(
       .setTimeout(0)
       .build()
     transaction.sign(keypair)
-    this.logger.request('Submitting changeTrust transaction', transaction)
+    this.logger.request(
+      'Submitting claimClaimableBalance transaction',
+      transaction
+    )
     const result = await this.server.submitTransaction(transaction)
     this.logger.response('Submitted claimClaimableBalance transaction', result)
     await this.updateAccount()
--- a/src/components/wallet/methods/createAccount.ts
+++ b/src/components/wallet/methods/createAccount.ts
@@ -19,7 +19,6 @@ export default async function createAccount(this: Wallet) {
     this.account = {
       publicKey: keypair.publicKey(),
       secretKey: keypair.secret(),
-      availableBalances: false,
     }
 
     set('WALLET[keystore]', btoa(JSON.stringify(this.account)))
--- a/src/components/wallet/methods/createAccount.ts
+++ b/src/components/wallet/methods/createAccount.ts
@@ -20,6 +20,7 @@ export default async function createAccount(this: Wallet) {
       publicKey: keypair.publicKey(),
       secretKey: keypair.secret(),
     }
+    this.assets.set('XLM', {})
 
     set('WALLET[keystore]', btoa(JSON.stringify(this.account)))
 
--- a/src/components/wallet/methods/createAccount.ts
+++ b/src/components/wallet/methods/createAccount.ts
@@ -19,6 +19,7 @@ export default async function createAccount(this: Wallet) {
     this.account = {
       publicKey: keypair.publicKey(),
       secretKey: keypair.secret(),
+      availableBalances: false,
     }
 
     set('WALLET[keystore]', btoa(JSON.stringify(this.account)))
--- a/src/components/wallet/views/claimableDisplay.tsx
+++ b/src/components/wallet/views/claimableDisplay.tsx
@@ -4,17 +4,15 @@ import WalletButton from './walletButton'
 import { Wallet } from '../wallet'
 
 interface ClaimableBalance {
-  id: string;
-  paging_token: string;
-  asset: string;
-  amount: string;
-  sponsor?: string;
-  last_modified_ledger: number;
-  claimants: any[];
+  id: string
+  asset: string
+  amount: string
+  sponsor?: string
+  claimants: any[]
 }
 
 export default function claimableDisplay(this: Wallet) {
-  if (this.account.availableBalances){
+  if (this.account.availableBalances) {
     const claimableRecords = (claimableBalances: any) => {
       return claimableBalances.records.map(balanceRow)
     }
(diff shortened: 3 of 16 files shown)


JakeUrban:
I haven't looked at the code yet but I just wanted to suggest a small UI addition:
Screen Shot 2020-11-12 at 9 58 17 AM

Can we have some sort of `Claimable Balances` header above the added asset balance view?

JakeUrban:
Although I just realized that if we add a header to the Claimable Balances view we should also do so for the normal balances. Maybe `Claimable Balances` and `Balances`?

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+            <div class="balance">{balance.amount}</div>
+          </div>
+          <div class="actions">{claimBalanceButton}</div>
+        </div>
+      )
+    }
+
+    return (
+      <div>
+        {loHas(this.account, 'claimableBalances') ? (
```
JakeUrban, line 50: What is `loHas`? I've never understood the `lo*` functions

src/components/wallet/methods/createAccount.ts
```diff
@@ -19,6 +19,7 @@ export default async function createAccount(this: Wallet) {
     this.account = {
       publicKey: keypair.publicKey(),
       secretKey: keypair.secret(),
+      availableBalances: false,
```
JakeUrban, line 22: I'm not against having this attribute if its a pain to determine it's value from `ServerApi.ClaimableBalanceRecord`, but it seems to just be a matter of checking an array length.

In which case, I think having an `availableBalances` attribute actually does more harm than good because `availableBalances`'s value and the real `claimableBalanceResp.records.length === 0` might not match if we make a change in future and forget to update `availableBalances`.

src/components/wallet/wallet.ts
```diff
@@ -51,6 +53,7 @@ interface Loading {
   update?: boolean
   deposit?: boolean // NEW
   withdraw?: boolean // NEW
+  claim?: boolean // NEW
```
JakeUrban, line 56: I think we can remove the `// NEW` comments scattered throughout the project, they don't provide any value

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+interface ClaimableBalance {
+  id: string
+  asset: string
+  amount: string
+  sponsor?: string
+  claimants: any[]
+}
+
+export default function claimableDisplay(this: Wallet) {
+  if (this.account.availableBalances) {
```
JakeUrban, line 15: I'm not sure if this is the right code (is there a difference between `this.account.claimableBalances` and `claimableBalanceResp` used to populate `this.account.availableBalances`?), but you get the gist of the suggestion:
```
  if (this.account.claimableBalances.records.length !== 0) {
```

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+    const claimableRecords = (claimableBalances: any) => {
```
JakeUrban, line 16: No worries if not but are we able to use a proper type or no? Is an item in `claimableBalances` a `claimableBalanceRecord`?

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+      return claimableBalances.records.map(claimableBalanceRow)
+    }
+    const claimableBalanceRow = (balance: ClaimableBalance) => {
```
JakeUrban, line 19: Can we move this function out of the outer function for readability?

src/components/wallet/views/claimableDisplay.tsx
JakeUrban: And when is `this.account.claimableBalances` populated?

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,58 @@
...
+      const loadingKey = (type: string) => {
+        return `${type}:${balance.asset.split(':')[0]}:${balance.id}`
+      }
+      const assetCode = balance.asset.split(':')[0]
+      const claimBalanceButton = WalletButton.call(
+        this,
+        'Claim',
+        loadingKey('claim'),
+        () => {
+          this.claimAsset(
```
JakeUrban, line 29: It seems to me that we could get away with only passing `balance`, `this.claimAsset(balance)`?

Zagan202:
> Although I just realized that if we add a header to the Claimable Balances view we should also do so for the normal balances. Maybe `Claimable Balances` and `Balances`?

Thats totally doable give me a few min and I'll push a change

src/components/wallet/views/claimableDisplay.tsx
JakeUrban: Oh actually it looks like there are 3 function layers here, `claimableDisplay`, `claimableRecords`, and `claimableBalanceRow`. We should at least move `claimableRecords` outside `claimableDisplay`.

src/components/wallet/methods/claimAsset.ts
```diff
@@ -0,0 +1,63 @@
...
+export default async function claimAsset(
+  this: Wallet,
+  balanceId: string,
+  assetCode: string,
+  sponsor: string,
+  amount: string
+) {
+  const loadingKey = `claim:${assetCode}:${balanceId}`
+  this.loading = { ...this.loading, [loadingKey]: true }
+  const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
```
JakeUrban, line 21: Interesting approach. Why does the key need to be an array of size one? And where is this key-value pair used?

JakeUrban (changes requested):
Great work, many of my comments are just questions so I can understand. I do have some suggestions though.

src/components/wallet/views/claimableDisplay.tsx
Zagan202: the lo* functions are used to check, manipulate, etc (ie the [lodash](https://lodash.com/) functions) are used to make it "easier" to interface with json items primarily in this codebase.
It is nice since it will look for really nested keys.
so `loHas(this.account, 'claimableBalances')` is a way to check the existence of a claimableBalances key

Zagan202: `claimableBalances` is populated updated etc in during the updateAccount process.
Which happens frequently and is populated before claimable balances are displayed.

src/components/wallet/methods/createAccount.ts
Zagan202: I'm not a big fan of this implementation tbh but I think I'm going to address this a bit more elegantly in a future PR.

src/components/wallet/wallet.ts
Zagan202: Good point

src/components/wallet/methods/createAccount.ts
JakeUrban: Is there a reason we have to keep it as-is for now? It seems to be a easy change to remove the attribute and replaces uses of it with `this.account.claimableBalances.records.length`.

If you have an even better solution I'm curious to hear it

JakeUrban:
I think I found a small bug:

Screen Shot 2020-11-12 at 12 06 03 PM

JakeUrban:
One more change: can we change the `Update Account` button to have `Refresh Account` text?

Zagan202:
> I think I found a small bug:
> 
> Screen Shot 2020-11-12 at 12 06 03 PM

How can I reproduce this bug?

JakeUrban:
> How can I reproduce this bug?

Create an account and send an XLM claimable balance to it.

Zagan202:
> One more change: can we change the `Update Account` button to have `Refresh Account` text?

Sure thats no problem.

Zagan202:
> > How can I reproduce this bug?
> 
> Create an account and send an XLM claimable balance to it.

Ah it looks like Horizon returns back that result as `Native`. 
Not a problem intercepting that and rename it XLM. Similar to how the Balances section does it.

Zagan202:
> Oh actually it looks like there are 3 function layers here, `claimableDisplay`, `claimableRecords`, and `claimableBalanceRow`. We should at least move `claimableRecords` outside `claimableDisplay`.

Refactoring this is a bit more confusing and cumbersome than anticipated so I'm going to keep it the way it is.
It's also similar in style to the how `balanceDisplay.tsx` is written. 

> It seems to me that we could get away with only passing `balance`, `this.claimAsset(balance)`?

The way `claimAsset` is called its easier to split these parameters separately.
this.account.claimableBalances is an array of claimable balance records iterated through mapping helper func `claimableRecords` we need `this` to invoke this objects server but we need to specify which CB we want to claim ie(`id, asset, balance, and sponsor`)

 > Interesting approach. Why does the key need to be an array of size one? And where is this key-value pair used?

This is the current approach used elsewhere to invoke the loading animation.

JakeUrban:
> It's also similar in style to the how balanceDisplay.tsx is written.

Is it? I don't see any nested functions in the current [balanceDisplay.tsx](https://github.com/stellar/stellar-demo-wallet/blob/master/src/components/wallet/views/balanceDisplay.tsx)

> we need this to invoke this objects server but we need to specify which CB we want to claim ie(id, asset, balance, and sponsor)

Take a look at the code again. Everything you pass to `claimAsset()` is derived from the `balance` on each iteration, so you don't need to pass the exact properties from `balance` like `id`, `asset`, etc. because you can just access the same properties from within the `claimAsset()` function. `this` is automatically added to all object functions, just like `self` for python.

JakeUrban:
> This is the current approach used elsewhere to invoke the loading animation.

Can you point to the file does it this way?

src/components/wallet/views/claimableDisplay.tsx
```diff
@@ -0,0 +1,60 @@
...
+            balance.id,
+            assetCode,
+            balance.sponsor,
+            balance.amount
+          )
```
JakeUrban, line 35:
```suggestion
          this.claimAsset(balance)
```

src/components/wallet/methods/claimAsset.ts
```diff
@@ -0,0 +1,66 @@
...
+} from 'stellar-sdk'
+import { handleError } from '@services/error'
+import { Wallet } from '../wallet'
+
+export default async function claimAsset(
+  this: Wallet,
+  balanceId: string,
+  assetCode: string,
+  sponsor: string,
+  amount: string
```
JakeUrban, line 17:
```suggestion
export default async function claimAsset(
  this: Wallet,
  balance: any
```

src/components/wallet/methods/claimAsset.ts
```diff
@@ -0,0 +1,66 @@
...
+) {
+  const loadingKey = `claim:${assetCode}:${balanceId}`
```
JakeUrban, line 19:
```suggestion
  const assetCode = balance.asset === 'native' ? 'XLM' : balance.asset.split(':')[0]
  const loadingKey = `claim:${assetCode}:${balance.id}`
```

src/components/wallet/methods/claimAsset.ts
```diff
@@ -0,0 +1,66 @@
...
+  this.loading = { ...this.loading, [loadingKey]: true }
+  const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
+
+  try {
+    this.logger.instruction(
+      `Claiming ${amount} of ${assetCode}`,
+      `BalanceId: ${balanceId} Sponsor:${sponsor}`
```
JakeUrban, line 26:
```suggestion
      `Claiming ${balance.amount} of ${assetCode}`,
      `BalanceId: ${balance.id} Sponsor: ${balance.sponsor}`
```

src/components/wallet/methods/claimAsset.ts
```diff
@@ -0,0 +1,66 @@
...
+    )
+    const account = new Account(keypair.publicKey(), accountRecord.sequence)
+    this.logger.instruction('Building claimClaimableBalance transaction')
+    const transaction = new TransactionBuilder(account, {
+      fee: BASE_FEE,
+      networkPassphrase: Networks.TESTNET,
+    })
+      .addOperation(
+        Operation.claimClaimableBalance({
+          balanceId,
```
JakeUrban, line 44:
```suggestion
          balance.id,
```

JakeUrban (changes requested):

src/components/wallet/methods/claimAsset.ts
Zagan202: > > This is the current approach used elsewhere to invoke the loading animation.
> 
> Can you point to the file does it this way?
```
/Users/lawrence/stellar-dev/stellar-demo-wallet/src/components/wallet/methods/depositAsset.ts
  14,1:   this.loading = { ...this.loading, [loadingKey]: true }
  15:   

/Users/lawrence/stellar-dev/stellar-demo-wallet/src/components/wallet/methods/makePayment.ts
  47,3:     this.loading = { ...this.loading, [loadingKey]: true }
  48:   

/Users/lawrence/stellar-dev/stellar-demo-wallet/src/components/wallet/methods/makeRegulatedPayment.tsx
  48,3:     this.loading = { ...this.loading, [loadingKey]: true }
  49:   

/Users/lawrence/stellar-dev/stellar-demo-wallet/src/components/wallet/methods/withdrawAsset.ts
  32,1:   this.loading = { ...this.loading, [loadingKey]: true }
  33:   
```

src/components/wallet/methods/claimAsset.ts
```diff
@@ -7,23 +7,23 @@ import {
...
 export default async function claimAsset(
   this: Wallet,
-  balanceId: string,
-  assetCode: string,
-  sponsor: string,
-  amount: string
+  balance: ClaimableBalance
 ) {
-  const loadingKey = `claim:${assetCode}:${balanceId}`
+  const balanceId = balance.id
```
JakeUrban, line 17: Is this variable used? I see you use `balance.id` everywhere else

JakeUrban (approved):
One comment because I think the TS compiler will complain about an unused variable. Otherwise LGTM!

src/components/wallet/methods/claimAsset.ts
Zagan202: For some reason ts was complaining when I used `balance.id` here `Operation.claimClaimableBalance`.
So I assigned it to a new variable `balanceId` where its called on line 44
```
      .addOperation(
        Operation.claimClaimableBalance({
          balanceId,
        })
      )
```

JakeUrban: Oh I see one usage. Ok! I'm going to merge this

JakeUrban: Ohh I bet it was because `claimClaimableBalance` expects an object with `balanceId` as a key. The `{ key }` shorthand syntax is short for: `{ key: value }` but it only works when the variable name matches the key name.

PR #52, opened 2020-11-13
Add secretKey URL param for loading existing accounts

### LINKED ISSUE #35: Make configurable
This demo wallet was originally designed as if it was going to be a real wallet. It doesn't provide easy configuration menus that make life much easier for developers building Stellar services and using this application to test. 

Furthermore, it is not possible to send a fellow dev a link that recreates the configuration state I had locally.

What we'll do:
- Provide a Config menu that updates the URL on change
- If not used, the demo wallet will present a prompt asking for the asset code and either the home domain or issuer address

JakeUrban:
resolves #35 

Functional Changes:
- If `secretKey` URL param is present, load that account
- Add/Remove `secretKey` URL param when creating/loading accounts or signing out
- Removes "Are you sure" prompt when signing out. Users can simply press the back button.
- No longer attempts to create the account from `loadAccount()`

Non-Functional Changes:
- Update dependencies (removed unneeded electron app and menubar package)
- Updated `lib` in `tsconfig.json`, allowing us to use ES2019 syntax


src/components/wallet/methods/signOut.ts
```diff
@@ -1,22 +1,13 @@
 import { remove } from '@services/storage'
 import { handleError } from '@services/error'
 import { Wallet } from '../wallet'
-import { PromptInput } from '../../prompt/promptInput'
 
 export default async function signOut(this: Wallet) {
   try {
-    const inputs = await this.setPrompt({
```
Zagan202, line 8: IMO I don't think we should remove the confirmation prompt entirely.
It could be better to have a prompt that gives the user a warning and an opportunity to save their secretkey to a clipboard before their account gets _"nuked"_.

That being said remove the `"Enter NUKE to confirm"` part
but include a press enter to confirm
and add `copy secret key` to the same prompt.

Zagan202 (changes requested):

src/components/wallet/methods/signOut.ts
JakeUrban: I replaced the text message with "You can press the back button to sign back in.", but right now we can't pass buttons to the prompt for copying the secret key. The secret key is also in the URL now so the button isn't as needed. LMK if this works.

src/components/wallet/methods/signOut.ts
```diff
@@ -1,22 +1,19 @@
...
-import { PromptInput } from '../../prompt/promptInput'
+import { PromptInput } from '@prompt/promptInput'
 
 export default async function signOut(this: Wallet) {
   try {
-    const inputs = await this.setPrompt({
-      message: 'Are you sure? This will nuke your account',
-      inputs: [new PromptInput('Enter NUKE to confirm')],
+    await this.setPrompt({
+      message: 'You can press the back button to sign back in.',
```
Zagan202, line 8: Minor messaging suggestion.

```suggestion
      message: 'You can enter your Secret Key or press back in your browser to sign back in.',
```

Zagan202 (approved):
Besides the one suggestion it LGTM

src/components/wallet/methods/signOut.ts
JakeUrban:
```suggestion
      message: 'You can add your Secret Key to the URL or press back in your browser to sign back in.',
```

Zagan202: Looks good to me!

Issue #53, opened 2020-11-14
Add Mainnet Support


JakeUrban:
The old demo client can run everything on mainnet, we should offer the same functionality on the demo wallet.

This should simply involve making the `NETWORK_PASSPHRASE` used configurable, with some kind of toggle button for switching between testnet and mainnet.

The only challenge I see with this issue is this project's use of Friendbot for `createAccount()`. We may have to remove the `Create Account` button if mainnet is selected.

leighmcculloch:
➕ 1️⃣  although we should refer to it as Pubnet, as that's what the public Stellar network is called in other tools and documentation.

JakeUrban:
Oh thats good to know, we definitely call it mainnet in all our integration resources sent to partners. I always assumed the two words could be used interchangeably. I almost prefer 'mainnet' because 'pubnet' implies the other network would be 'privatenet'

leighmcculloch:
I don't have a personal preference other then we should be consistent and it's not worth changing. Most public documentation, tools, and SDKs use the term testnet/pubnet or test/public.

Examples:

- https://www.stellar.org/developers/guides/concepts/networks.html#network-passphrases
   >Pubnet: `'Public Global Stellar Network ; September 2015'`
   >Testnet: `'Test SDF Network ; September 2015'`

- https://stellar.github.io/js-stellar-sdk/global.html#Networks

- https://stellar.expert/explorer/public

PR #55, opened 2020-11-18
Add Support for Pubnet

JakeUrban:
resolves stellar/stellar-demo-wallet#53 

Adds a new button to the UI that allows the user to toggle between pubnet and testnet. Defaults to testnet.

Under the hood, clicking this button essentially signs you out (like `signOut()`) and reloads the application with the URL param `pubnet=true` either present or not present. When reloading, `componentWillLoad()` determines what the value of `Wallet.server` and `Wallet.network_passphrase` need to be.

Note that the `Create Account` button is not present when using pubnet. This is because we don't have friendbot on pubnet. We could eventually add functionality to allow the user to create an account from an existing account on mainnet, but for now its not needed.

### DIFF
--- a/src/components/wallet/events/componentWillLoad.ts
+++ b/src/components/wallet/events/componentWillLoad.ts
@@ -1,7 +1,7 @@
 import { handleError } from '@services/error'
 import { get } from '@services/storage'
 import { Wallet } from '../wallet'
-import { Keypair } from 'stellar-sdk'
+import { Keypair, Networks, Server } from 'stellar-sdk'
 
 export default async function componentWillLoad(this: Wallet) {
   try {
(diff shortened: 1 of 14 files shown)


JakeUrban:
@leighmcculloch your review is optional here since you were curious about this task. @Zagan202 your review is not optional :-)

src/components/wallet/events/componentWillLoad.ts
```diff
@@ -18,6 +18,13 @@ async function loadAccountFromBrowser(wallet: Wallet) {
   } catch {
     throw 'Unable to parse query string'
   }
+  if (query.pubnet === 'true') {
+    wallet.network_passphrase = Networks.PUBLIC
+    wallet.server = new Server('https://horizon.stellar.org')
```
Zagan202, line 23: In a separate PR I think we should replace these url strings with a global constant like `TESTNET_URI` and `MAINNET_URI`.

Zagan202 (approved):
LGTM

PR #57, opened 2020-11-21
Replace http with https on interactiveResponse and more_info_url

Zagan202:
# Description

In order to debug the demo wallet that is attempting to deposit on a local instance of Polaris (or any other anchor service ran locally).

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?
 
Depositing fake "offchain assets" to a local instance of Polaris with an SSL proxy (in order to emulate more of a production like behavior)

Signed-off-by: Lawrence Lawson <lawrence@stellar.org>

### DIFF
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -163,8 +163,9 @@ export default async function depositAsset(
     if (!interactiveResponse.url) {
       throw 'No URL Returned from POST /transactions/deposit/interactive'
     }
-
-    const urlBuilder = new URL(interactiveResponse.url)
+    const urlBuilder = new URL(
+      interactiveResponse.url.replace(/^http:\/\//i, 'https://')
+    )
     urlBuilder.searchParams.set('callback', 'postMessage')
     this.logger.instruction(
       'To collect the interactive information we launch the interactive URL in a frame or webview, and await payment details from a postMessage callback'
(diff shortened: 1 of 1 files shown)


src/components/wallet/methods/depositAsset.ts
```diff
@@ -163,8 +163,9 @@ export default async function depositAsset(
     if (!interactiveResponse.url) {
       throw 'No URL Returned from POST /transactions/deposit/interactive'
     }
-
-    const urlBuilder = new URL(interactiveResponse.url)
+    const urlBuilder = new URL(
+      interactiveResponse.url.replace(/^http:\/\//i, 'https://')
+    )
```
leighmcculloch, line 168: 💡 Instead of breaking out the big guns regex you can change the protocol directly on the `urlBuilder` object.
```suggestion
    const urlBuilder = new URL(interactiveResponse.url)
    urlBuilder.protocol = 'https'
```

src/components/wallet/methods/depositAsset.ts
```diff
@@ -183,7 +184,9 @@ export default async function depositAsset(
       } else {
         this.logger.instruction('Transaction status pending...')
         setTimeout(() => {
-          const urlBuilder = new URL(transaction.more_info_url)
+          const urlBuilder = new URL(
+            transaction.more_info_url.replace(/^http:\/\//i, 'https://')
```
leighmcculloch, line 188: 💡 Same as above.

leighmcculloch (approved):
Nice move to default to https. Suggestion (💡) below, otherwise LGTM.

src/components/wallet/methods/depositAsset.ts
Zagan202: Oh wow thats a lot easier 😅 
Thanks for the suggestion.

JakeUrban:
Sorry I'm just seeing this. It looks like this isn't "defaulting" to HTTPS, its forcing it. If I'm running an HTTP anchor locally, this will break. @Zagan202 can you explain your reasoning here?

PR #60, opened 2020-11-25
Update strict homedomain protocol requirements

Zagan202:
# Description

This fixes a blocker I introduced in https://github.com/stellar/stellar-demo-wallet/pull/57 where we force the protocols of the urls we require to be `https`.
This made it impossible to connect a local anchor server(polaris, etc.) to the `stellar-demo-wallet` unless the dev tunneled their local anchor server through a SSL proxy and passed the https url and port.

The following change does not enforce this; instead it will accept http or https protocols and attempt to fetch data from the given location. However if no protocol is given it will assume (and then append) `https://`

### Original Issue Context

In the past whenever I attempted to connect an anchor service to the `stellar-demo-wallet` to a local Polaris server it would fail to fetch the TOML when trusting a dummy asset.

This blocked me from demoing a deposit flow since I could not trust the asset.

The solution was
- Ensure protocols were consistent, meaning `http` stayed `http` and `https` stayed `https`
- Invoke StellarTomlResolver.resolve() correctly; `StellarTomlResolver.resolve(tomlURL.host, { allowHttp: true })` for `http` `StellarTomlResolver.resolve(tomlURL.host)` for `https`

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

1. Run `Polaris` locally with `poll_pending_deposits`
1. Create an account 
1.  Deposit deployed assets with https
     - Trust SRT https://testanchor.stellar.org
    - Deposit $10
    - Expected to pass
1. Deposit local assets with http
    - Trust SRT or any other dummy asset http://localhost:8000/
    - Deposit $10
    - Expected to pass
1. Deposit deployed assets without https
    - Trust SRT https://testanchor.stellar.org
    - Deposit $10
    - Expected to pass
1. Deposit local assets without http
    Trust SRT or any other dummy asset localhost:8000/
    - Expected to fail
        - This is by design since it assumes its `https` and appends `https://` on a domain that is is http
        - Then when fetching the `toml` it fails
    ```markdown
    Error in trust transaction
    {
    config: 12 values
    {
    url: https://localhost:8000/.well-known/stellar.toml
    method: get
    headers: 1 values
    transformRequest: 1 values
    transformResponse: 1 values
    timeout: 0
    adapter: Function
    xsrfCookieName: XSRF-TOKEN
    xsrfHeaderName: X-XSRF-TOKEN
    maxContentLength: 102400
    validateStatus: Function
    data: undefined
    }
    request: 0 values
    response: undefined
    isAxiosError: true
    toJSON: Function
    }
    ```

### DIFF
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -44,10 +44,9 @@ export default async function depositAsset(
       }
       homeDomain = inputs[0].value
     }
-    homeDomain =
-      homeDomain.startsWith('http://') || homeDomain.startsWith('https://')
-        ? homeDomain
-        : 'https://' + homeDomain
+    homeDomain = homeDomain.startsWith('http')
+      ? homeDomain
+      : 'https://' + homeDomain
     const tomlURL = new URL(homeDomain)
     tomlURL.pathname = '/.well-known/stellar.toml'
     this.logger.request(tomlURL.toString())
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -44,9 +44,10 @@ export default async function depositAsset(
       }
       homeDomain = inputs[0].value
     }
-    homeDomain = (homeDomain.startsWith('http://') || homeDomain.startsWith('https://'))
-      ? homeDomain
-      : 'https://' + homeDomain
+    homeDomain =
+      homeDomain.startsWith('http://') || homeDomain.startsWith('https://')
+        ? homeDomain
+        : 'https://' + homeDomain
     const tomlURL = new URL(homeDomain)
     tomlURL.pathname = '/.well-known/stellar.toml'
     this.logger.request(tomlURL.toString())
--- a/src/components/wallet/methods/depositAsset.ts
+++ b/src/components/wallet/methods/depositAsset.ts
@@ -51,7 +51,10 @@ export default async function depositAsset(
     const tomlURL = new URL(homeDomain)
     tomlURL.pathname = '/.well-known/stellar.toml'
     this.logger.request(tomlURL.toString())
-    const toml = await StellarTomlResolver.resolve(tomlURL.host)
+    const toml =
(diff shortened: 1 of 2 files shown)


src/components/wallet/methods/depositAsset.ts
```diff
@@ -44,14 +44,17 @@ export default async function depositAsset(
       }
       homeDomain = inputs[0].value
     }
-
-    homeDomain = homeDomain.startsWith('https://')
-      ? homeDomain
-      : 'https://' + homeDomain
+    homeDomain =
+      homeDomain.startsWith('http://') || homeDomain.startsWith('https://')
```
JakeUrban, line 48: This is explicit so we could keep it, but you could just do:
```suggestion
      homeDomain.startsWith('http')
```
Since this will return `true` for domains with `https` and `http`.

src/components/wallet/methods/depositAsset.ts
```diff
@@ -44,14 +44,17 @@ export default async function depositAsset(
...
+        ? homeDomain
+        : 'https://' + homeDomain
     const tomlURL = new URL(homeDomain)
     tomlURL.pathname = '/.well-known/stellar.toml'
     this.logger.request(tomlURL.toString())
-    const toml = await StellarTomlResolver.resolve(tomlURL.host)
+    const toml =
+      tomlURL.protocol == 'http:'
+        ? await StellarTomlResolver.resolve(tomlURL.host, { allowHttp: true })
```
JakeUrban, line 56: nice

JakeUrban (approved):
LGTM, one suggestion

src/components/wallet/methods/depositAsset.ts
Zagan202: Good catch!

abuiles:
btw, I was thinking about this -- maybe a better approach is not to do this (which is kind of brittle)  but rather always develop locally with https? that way it is closer to what you expect in prod -- this can be done very easily with tools like https://github.com/octavore/lightproxy

Issue #64, opened 2020-12-16
Feature Request: Deposit Untrusted Assets


Zagan202:
### What problem does your feature solve?

Leveraging the claimable balance feature in SEP24 to allow accounts to deposit off-chain assets to anchors without a trustline.
Developer should be able to debug the scenario  where they add the asset in the UI and go through a deposit flow without adding trustlines.

### What would you like to see?

-  “Add asset” button enters the issuer/home domain _but_ skips adding trust to said asset .
- _Add an_ “add trust” button that allows them to add trustline explicitly. 
- _Behind the scenes_, when claiming balances from those assets add a trustline to account (one time) to claim.
- _Remove_ "untrusted" label when trustline is established.
- ~~List a select few assets on the wallet's account page~~ 
- ~~Label the assets that qualify as "untrusted"~~
- ~~Allow users to make deposits on those listed untrusted assets~~

### What alternatives are there?

https://github.com/Zagan202/stellar-demo-wallet/tree/deposit-untrusted-TEST-asset

PR #65, opened 2020-12-21
Deposit Untrusted Assets

Zagan202:
# Description

Allow wallet users to deposit off-chain assets to anchors without a trustline.

- [X] “Add Asset” button enters the issuer/home domain but skips adding trust to said asset .
- [X] Add an “Add Trust” button that allows users to add trustline explicitly.
- [X] Claiming balances from those assets adds trustline to account (one time) to claim.
- [X] Added and untrusted assets in `Balances` can only be add a trustline and Deposit to a claimable balance supported anchor
- [X] When Asset is added `Trust Asset` button appears when asset is trusted it is not and we see the typical `Send Deposit Withdraw` buttons.

Addresses issue: #64 

## Type of change

# How Has This Been Tested?

## Deposit Untrusted Asset
1. `yarn install` `yarn start`
2. go to `http://localhost:3333/`
3. `Create Account`
4. `Add Asset`
    - asset code: `SRT` home domain: `https://testanchor.stellar.org/`
    - You should see this in your `Balances `
    ```
     SRT:
     0.0000000
     TrustAsset Deposit
     ```
5.  Attempt a deposit with the deployed testanchor
    - You can insert any dummy information
    - The Deposit may error out at this time due to some high DEX traffic
6. If Deposit is successful you should see that amount deposited in your Claimable Balances
7. Attempt to claim your claimable balance
    - The claim may error out at this time due to some high DEX traffic
8. If successful your Balance should look like 
    ```
     SRT:
     10.0000000
     Send Deposit Withdraw
     ```

##  Trust Asset
1. `yarn install` `yarn start`
2. go to `http://localhost:3333/`
3. `Create Account`
4. `Add Asset`
    - asset code: `SRT` home domain: `https://testanchor.stellar.org/`
    - You should see this in your `Balances `
    ```
     SRT:
     0.0000000
     Trust Asset Deposit
     ```
5. Attempt to trust asset
    - Adding trustline may error out at this time due to some high DEX traffic
8. If successful your Balance should look like 
    ```
     SRT:
     0.0000000
     Send Deposit Withdraw
     ```
##  Claimable Balance Untrusted Asset (From Fresh State)
1. `yarn install` `yarn start`
2. go to `http://localhost:3333/`
3. `Create Account`
4. `Add Asset`
    - asset code: `SRT` home domain: `https://testanchor.stellar.org/`
    - You should see this in your `Balances `
    ```
     SRT:
     0.0000000
     Trust Asset Deposit
     ```
5.  Attempt a deposit with the deployed testanchor
    - You can insert any dummy information
    - The Deposit may error out at this time due to some high DEX traffic
6. If Deposit is successful you should see that amount deposited in your Claimable Balances
7. Save Secret Key and sign out and sign in (removing your state on storage)
8. Verify claimable balance is still present but `Balances` will only include assets that are trusted (XLM etc)
7. Attempt to claim your claimable balance
    - The claim may error out at this time due to some high DEX traffic
8. If successful your Balance should look like 
    ```
     SRT:
     10.0000000
     Send Deposit Withdraw
     ```


src/components/wallet/methods/updateAccount.ts
```diff
@@ -30,13 +31,37 @@ export default async function updateAccount(this: Wallet) {
...
     }
+    const balKeystore = await get('BALANCE[keystore]')
+    // this restores the balance prop from storage
+    if (balKeystore && !this.balance.size) {
+      this.balance = new Map()
+      JSON.parse(atob(balKeystore)).forEach((b) => {
+        this.balance.set(b[0], b[1])
+      })
+    }
     account.balances.forEach((b) => {
```
JakeUrban, line 42: I think you need to `set('BALANCE[keystore]', ...)` if there are any new assets in the account right?

What I would do is merge `account.balances` with `this.balance` so that `this.balance` is up to date with whats in the Stellar account PLUS the balances that don't have a trustline.

src/components/wallet/views/balanceDisplay.tsx
```diff
@@ -18,12 +10,15 @@ export default function balanceDisplay(this: Wallet) {
     }
     const isRegulated =
       !balance.is_authorized && balance.asset_type !== 'native'
+    const isNotTrusted = !balance.trusted && balance.asset_type !== 'native'
```
JakeUrban, line 13: I think you can just use `!balance.trusted` because you have XLM marked as `trusted: true`

src/components/wallet/methods/addAsset.ts
```diff
@@ -0,0 +1,53 @@
...
+        asset_type: addAsset.asset_type,
+        is_authorized: true,
+        trusted: false,
+      })
+      set(
+        'BALANCE[keystore]',
+        btoa(JSON.stringify(Array.from(this.balance.entries())))
+      )
+    }
+    await this.updateAccount()
```
JakeUrban, line 45: I actually don't think this needs to be called here either, because the account state didn't change and we already updated `this.balance`, right?

src/components/wallet/methods/claimAsset.ts
```diff
@@ -20,6 +20,11 @@ export default async function claimAsset(
   this.loading = { ...this.loading, [loadingKey]: true }
   const finish = () => (this.loading = { ...this.loading, [loadingKey]: false })
   try {
+    if (!this.balance.get(balance.asset).trusted) {
+      let asset = this.balance.get(balance.asset).asset_code
+      let issuer = this.balance.get(balance.asset).asset_issuer
```
JakeUrban, line 25: Just to make sure I understand correctly, these two lines could be replaced with:
```
let [asset, issuer] = balance.asset.split(":")
```
right?

src/components/wallet/methods/trustAsset.ts
```diff
@@ -115,6 +48,42 @@ export default async function trustAsset(
...
     this.logger.response('Submitted changeTrust transaction', result)
+    // update the balance prop with the new asset we added
+    if (this.balance.has(`${asset}:${issuer}`)) {
+      this.balance.delete(`${asset}:${issuer}`)
+    }
+    let loadedAccount = await this.server
+      .accounts()
+      .accountId(this.account.publicKey)
+      .call()
+    loadedAccount.balances.forEach((b) => {
```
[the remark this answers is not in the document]
JakeUrban: And then `updateAccount()` updates `this.assets`

JakeUrban (changes requested):
I requested some changes but overall nice job.

src/components/wallet/wallet.ts
```diff
@@ -38,6 +39,15 @@ const MockLogger = {
   },
 }
 
+export interface Balance {
+  balance: string
+  is_authorized: boolean
```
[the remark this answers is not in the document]
Zagan202: Done https://github.com/stellar/stellar-demo-wallet/issues/72

src/components/wallet/methods/claimAsset.ts
Zagan202: Oh great point!

src/components/wallet/methods/depositAsset.ts
```diff
@@ -18,6 +19,11 @@ export default async function depositAsset(
     // If the asset exists in the wallet UI, there exists a record in this.assets
     // however, the homeDomain may not be defined if a trustline for that asset
     // wasn't added in the same session
+    // If there is no trustline established then we'll ask for the
+    // Assetcode and homedomain in order to make a Claimable Balance Deposit
+    if (!this.assets.get(`${asset_code}:${asset_issuer}`)) {
+      await getAssetAndIssuer(this)
+    }
```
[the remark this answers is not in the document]
Zagan202: The primary reason was to ensure there was a homeDomain populated in order to initialize the deposit flow.
Without getting into too much in the weeds with the state management of wallet attributes; I can make this work as I think you intend it to via this code change 
```suggestion
    // If there is no trustline established then we'll ask for the
    // Assetcode and homedomain in order to make a Claimable Balance Deposit
    let homeDomain = null
    if (this.assets.get(`${asset_code}:${asset_issuer}`)){
      homeDomain = this.assets.get(`${asset_code}:${asset_issuer}`).homeDomain
    }
```
Note: This check needs to be made in order to not fail when attempting to access an attribute to a key value that isn't present (due to not saving the Wallet's asset property to disk). 

However another scenario that I think is an argument for keeping the `getAssetAndIssuer` call there is if you're debugging using an asset that doesn't have a home domain in the issuer's account.

For example, I issued a token on testnet named `LAWR` https://laboratory.stellar.org/#explorer?resource=assets&endpoint=single&values=eyJhc3NldF9jb2RlIjoiTEFXUiJ9&network=test
 However I did not set the [issuer's homedomain](https://developers.stellar.org/docs/start/list-of-operations/#set-options) primarily because it was anchored on a local debug instance of polaris.
 
 So I have 3 different paths to tackle this
 
 1. Make the change above and not allow anyone who hasn't set their homeDomain to make a deposit in the demo wallet.
 2. Keep the code the same and periodically (when you refresh) prompt the user to enter a homedomain to make a deposit.
 3. Refactor the code so we save the whole Wallet obj to disk (I believe I attempted this at one point and it didn't work out as intended but I can give it another go)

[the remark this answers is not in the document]
JakeUrban: Lets use your suggested change. If the home domain isn't set on the account, we ask for it on line 43.

src/components/wallet/methods/trustAsset.ts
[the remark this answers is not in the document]
Zagan202: You are 100% correct.
Great suggestion.

src/components/wallet/methods/depositAsset.ts
```diff
@@ -18,7 +18,12 @@ export default async function depositAsset(
     // If the asset exists in the wallet UI, there exists a record in this.assets
     // however, the homeDomain may not be defined if a trustline for that asset
     // wasn't added in the same session
```
JakeUrban, line 20: Lets remove this comment since it no longer applies right, there may not be a key for the asset within this.assets?

Zagan202:
## Recent Changes
- Change `BALANCE[keystore]` to only include untrusted assets
    - Now called `UNTRUSTEDASSETS[keystore]`
- Refactor `balanceDisplay` to pull from both  UntrustedAssets and account.state.balances
    - NOTE: `Balance` interface used for `UntrustedAssets` entries needs a flag (`untrusted: boolean`) to indicate whether or not it's trusted during `balanceDisplay`'s `balanceRow` mapping. 
- Port `LoadAccount`'s secret key(prompt) logic to `updateAccount`
    - Deleted LoadAccount and only call `updateAccount`

src/components/wallet/methods/addAsset.ts
```diff
@@ -0,0 +1,63 @@
...
+          asset_type: addAsset.asset_type,
+          is_authorized: true,
+          untrusted: true,
+        }
+      )
+      let UNTRUSTEDASSETS = {}
+      this.UntrustedAssets.forEach((value, key) => {
+        UNTRUSTEDASSETS[key] = value
+      })
+      set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(UNTRUSTEDASSETS)))
```
JakeUrban, line 54: Can't you use do:
```
set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(this.UntrustedAssets)))
```
Because `UNTRUSTEDASSETS` will be the same as `this.UntrustedAssets` right?

JakeUrban: I see the same logic elsewhere, this question is for all instances of this

src/components/wallet/methods/updateAccount.ts
```diff
@@ -4,13 +4,47 @@ import {
...
+      let query = `?secretKey=${this.account.secretKey}`
+      if (this.network_passphrase === Networks.PUBLIC) {
+        query += `&pubnet=true`
+      }
+      history.replaceState(null, '', query)
+      set('WALLET[keystore]', btoa(JSON.stringify(this.account)))
+      set(
+        'UNTRUSTEDASSETS[keystore]',
+        btoa(JSON.stringify(this.UntrustedAssets))
+      )
```
JakeUrban, line 45: aha! So it does work haha. Yea just use `this.UntrustedAssets` every time instead of building the `UNTRUSTEDASSETS` map.

src/components/wallet/views/balanceDisplay.tsx
```diff
@@ -3,16 +3,22 @@ import { has as loHas } from 'lodash-es'
...
+  if (this.account.state) {
+    Array.from(this.account.state.balances).forEach((bal) => {
+      totalBalances.push(bal)
+    })
+  }
+  if (this.UntrustedAssets.values) {
+    Array.from(this.UntrustedAssets.values()).forEach((bal) => {
+      totalBalances.push(bal)
+    })
+  }
```
JakeUrban, line 20:
```suggestion
  if (this.account.state) {
    totalBalances = Array.from(this.account.state.balances)
  }
  if (this.UntrustedAssets.values) {
    totalBalances = totalBalances.concat(Array.from(this.UntrustedAssets.values()))
  }
```

src/components/wallet/views/balanceDisplay.tsx
```diff
@@ -3,16 +3,22 @@ import { has as loHas } from 'lodash-es'
...
-  asset_issuer: string
-}
-
 export default function balanceDisplay(this: Wallet) {
-  const balanceRow = (balance: Balance) => {
+  let totalBalances = []
+  // Combines the two arrays of different types
+  // to allow balanceRow to iterate and map
+  // their values in the view
+  if (this.account.state) {
```
JakeUrban, line 11: If this view is being rendered, won't `this.account.state` always be populated?

src/components/wallet/views/loggedInContent.tsx
```diff
@@ -29,14 +29,13 @@ export default function loggedInContent(this: Wallet) {
           : 'Use Testnet'}
       </button>
     </div>,
-
```
JakeUrban, line 32: lets keep this whitespace or remove the other newlines in the array to keep it consistent

JakeUrban (changes requested):
Well done. I left some minor comments but I tested this branch out and everything seems to be working as expected -- including the ability to receive a claimable balance from Polaris and claim it!

src/components/wallet/methods/updateAccount.ts
Zagan202: Unfortunately I can't avoid build the map each time because for whatever reason when javascript attempts to encode the objects attribute directly the result when decoded is an empty map; even after we have populated the map.

```
!!!ENCODING DIRECTLY!!!

set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(this.UntrustedAssets))) 
=  
e30=

!!!DECODING THE DIRECT ATTRIBUTE!!!

const TEST = await get('UNTRUSTEDASSETS[keystore]')
const mapTEST = new Map(
        Object.entries(JSON.parse(atob(TEST)))
        )
=
Map(0) {}[[Entries]]No propertiessize: (...)__proto__: Map
------------------------------------------------------------------
!!!ENCODING THE BUILT MAP!!!

let UNTRUSTEDASSETS = {}
      this.UntrustedAssets.forEach((value, key) => {
        UNTRUSTEDASSETS[key] = value
      })
set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(UNTRUSTEDASSETS)))
 =  eyJTUlQ6R0NETkpVQlFTWDdBSldMSkFDTUo3STRCQzNaNDdCUVVUTUhFSUNaTEU2TVU0S1FCUllHNUpZNkIiOnsiYXNzZXRfY29kZSI6IlNSVCIsImFzc2V0X2lzc3VlciI6IkdDRE5KVUJRU1g3QUpXTEpBQ01KN0k0QkMzWjQ3QlFVVE1IRUlDWkxFNk1VNEtRQlJZRzVKWTZCIiwiYmFsYW5jZSI6IjAuMDAwMDAwMCIsImFzc2V0X3R5cGUiOiJjcmVkaXRfYWxwaGFudW00IiwiaXNfYXV0aG9yaXplZCI6dHJ1ZSwidW50cnVzdGVkIjp0cnVlfX0=

!!!DECODING THE BUILT MAP!!!

const TEST2UNTRUSTEDASSETS = await get('UNTRUSTEDASSETS[keystore]')
const map2TESTUNTRUST = new Map(
        Object.entries(JSON.parse(atob(TEST2UNTRUSTEDASSETS)))
)
= Map(1) {"SRT:GCDNJUBQSX7AJWLJACMJ7I4BC3Z47BQUTMHEICZLE6MU4KQBRYG5JY6B" => {…}}[[Entries]]0: {"SRT:GCDNJUBQSX7AJWLJACMJ7I4BC3Z47BQUTMHEICZLE6MU4KQBRYG5JY6B" => Object}size: (...)__proto__: Map
```

Zagan202: I looked around but I didnt find another way to stringify the object's map attribute without decoupling it from object it self (ie. building it).

Frankly idk why this  works at all -> `set('WALLET[keystore]', btoa(JSON.stringify(this.account)))`

Maybe it doesnt, I would have to investigate further to see if it does or not (the system behaves as if it does...).

Zagan202: If you have any suggestions Im open to hearing them because it was a weird bug.

Zagan202: Technically to be more "correct" I should change this to 
```suggestion
      let UNTRUSTEDASSETS = {}
      this.UntrustedAssets.forEach((value, key) => {
         UNTRUSTEDASSETS[key] = value
      })
     set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(UNTRUSTEDASSETS)))
```
However since `set('UNTRUSTEDASSETS[keystore]', btoa(JSON.stringify(UNTRUSTEDASSETS)))` already makes an empty map in a stage where an empty map is required it could stay as is.

src/components/wallet/views/balanceDisplay.tsx
Zagan202: Funny enough no.
When you first create an account state isnt populated yet.
Heres the call stack (sumarized a lot)
```
balanceDisplay (stellar-wallet.entry.js:49759)
loggedInContent (stellar-wallet.entry.js:49849)
render (stellar-wallet.entry.js:49891)
callRender (index-9c8102b3.js:1415)
updateComponent (index-9c8102b3.js:1348)
createAccount (stellar-wallet.entry.js:49901)
onClick (stellar-wallet.entry.js:49872)
```

In create account we dont populate the state
```
this.account = {
      publicKey: keypair.publicKey(),
      secretKey: keypair.secret(),
    };
    this.assets.set('XLM', {});
    set1('WALLET[keystore]', btoa(JSON.stringify(this.account)));
    ```
    So when loggedin content calls `balanceDisplay.call(this)`
    it will panic if we dont make this check.
    Before we were able to avoid this check due to just calling the state directly here 
    ```
      return (
    <div>
      {loHas(this.account, 'state') ? (
        <pre class="account-state">
          <h2 class="balance-headers">Balances</h2>
          {(this.account.state.balances as Balance[]).map(balanceRow)}
        </pre>
      ) : null}
    </div>
  )
  ```
  Where if there is no state we simply return null.
  However since we're building a concatenated list from both trusted assets and untrusted before we reach the return call that `{loHas(this.account, 'state')` ternary op will not really matter (I guess we could refactor it out now that I think about it).

File: src/components/wallet/methods/updateAccount.ts (hunk 10 above)
Comment (JakeUrban) [reply src/components/wallet/methods/updateAccount.ts:45]: I'd like to figure this out, because I don't believe there's some innate problem with serializing `this.UntrustedAssets`. I tested this in a JS console and there doesn't seem to be any difference in serialization between a `Map()` and an `object`:
Screen Shot 2021-01-14 at 10 04 30 AM

Comment (JakeUrban) [reply src/components/wallet/methods/updateAccount.ts:45]: Is it possible that `UNTRUSTEDASSETS[keystore]` is being overwritten and thats why it's empty?

Comment (Zagan202) [reply src/components/wallet/methods/updateAccount.ts:45]: I don't think I'm following.
I changed the objects attribute to include the new asset.

```
this.UntrustedAssets.set(
        `${addAsset.asset_code}:${addAsset.asset_issuer}`,
        {
          asset_code: addAsset.asset_code,
          asset_issuer: addAsset.asset_issuer,
          balance: '0.0000000',
          asset_type: addAsset.asset_type,
          is_authorized: true,
          untrusted: true,
        }
      )
```
Then I attempt to overwrite `UNTRUSTEDASSETS[keystore]` with the new changes.
```
set(
        'UNTRUSTEDASSETS[keystore]',
        btoa(JSON.stringify(this.UntrustedAssets))
      )
```
However it doesnt serialize it properly. only giving me an empty map.

Comment (Zagan202) [reply src/components/wallet/methods/updateAccount.ts:45]: Perhaps its the way I'm setting it.
I could try the key value approach instead of `set`
```
this.UntrustedAssets[`${addAsset.asset_code}:${addAsset.asset_issuer}`] = {
          asset_code: addAsset.asset_code,
          asset_issuer: addAsset.asset_issuer,
          balance: '0.0000000',
          asset_type: addAsset.asset_type,
          is_authorized: true,
          untrusted: true,
        }
```
Worth a shot.

JakeUrban (approved):

PR #78, opened 2021-01-21
Non development env should have hashed file names

ankurp:
Since the assets in upper environments do not change often it is best practice to hash the file names rather than have a simple file name as browser may cache and upon deploying newer changes may not be visible.

image


leighmcculloch (approved):

ankurp:
Sorry it did not work as expected. Need to check locally. Closing for now and will reopen.

PR #102, opened 2021-03-01
New UI balances and assets

quietbits:
- [x] Actions dropdown (from asset's TOML file)
- [x] Confirmation modal
- [x] Updated UI
- [x] Show toast banner when action is in progress
- [x] Refactor Asset type to be more consistent
- [ ] ~Add ability to cancel pending action (deposit or withdrawal when polling for a status change)~ Won't do right now, closing SEP-24 deposit and withdrawal popup will cancel the action.

### DIFF
--- a/src/components/Balance.tsx
+++ b/src/components/Balance.tsx
@@ -4,7 +4,12 @@ import { depositAssetAction } from "ducks/sep24DepositAsset";
 import { fetchSendFieldsAction } from "ducks/sep31Send";
 import { withdrawAssetAction } from "ducks/sep24WithdrawAsset";
 import { useRedux } from "hooks/useRedux";
-import { Asset, AssetActionItem, AssetActionId } from "types/types.d";
+import {
+  Asset,
+  AssetActionItem,
+  AssetActionId,
+  AssetType,
+} from "types/types.d";
 
 interface SortedBalancesResult {
   native: Asset[];
@@ -40,7 +45,7 @@ export const Balance = ({
     };
 
     Object.values(allBalances).map((balance) => {
-      if (balance.assetType === "native") {
+      if (balance.assetType === AssetType.NATIVE) {
         result.native = [...result.native, balance];
       } else {
         result.other = [...result.other, balance];
@@ -91,13 +96,16 @@ export const Balance = ({
     }
 
     let props: AssetActionItem | undefined;
+    const defaultProps = {
+      id: balance.assetString,
+      balance,
+    };
 
     switch (actionId) {
       case AssetActionId.SEND_PAYMENT:
         // TODO: title + description
         props = {
-          id: balance.assetString,
-          balance,
+          ...defaultProps,
           title: `Send payment ${balance.assetCode}`,
           description: "Send payment description",
           callback: onSend,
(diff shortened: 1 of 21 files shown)


src/components/Balance.tsx
```diff
@@ -44,35 +52,99 @@ export const Balance = ({
...
+    }
+
+    let props: AssetActionItem | undefined;
+
+    switch (actionId) {
+      case AssetActionId.SEND_PAYMENT:
+        // TODO: title + description
+        props = {
+          id: balance.assetString,
+          balance,
```
piyalbasu, line 100: very small - but would it make sense to declared a `defaultProps` object that = `{ id: balance.Assetstring, balance }` since every case where you're reassigning props, those are included?

For ex: 
```
let props: AssetActionItem | undefined;
const defaultProps = {
  id: balance.assetString,
  balance,
};

    switch (actionId) {
      case AssetActionId.SEND_PAYMENT:
        // TODO: title + description
        props = {
          ...defaultProps,
          title: `Send payment ${balance.assetCode}`,
          description: "Send payment description",
          callback: onSend,
        }

```

Would be helpful if in the future every case needs a new prop or one of these default props needs to be changed in some way

src/components/Balance.tsx
```diff
@@ -32,10 +40,10 @@ export const Balance = ({
     };
 
     Object.values(allBalances).map((balance) => {
-      if (balance.token.type === "native") {
-        result.native = [...result.native, balance as Types.NativeBalance];
+      if (balance.assetType === "native") {
```
piyalbasu, line 43: would it be helpful to have 
```
enum ASSET_TYPE {
  native = "native"
}
```
and check the equality against ASSET_TYPE.native instead? I see you're checking if something is `=== "native"` a few times.

src/components/UntrustedBalance.tsx
```diff
@@ -39,26 +56,77 @@ export const UntrustedBalance = () => {
...
+          title: `SEP-24 deposit ${asset.assetCode}`,
+          description: "Untrusted asset SEP-24 deposit description",
+          callback: () => handleDepositAsset(asset),
+        };
+        break;
+      case AssetActionId.TRUST_ASSET:
+        // TODO: title + description
+        props = {
+          id: asset.assetString,
+          balance: asset,
```
piyalbasu, line 87: similar case as `defaultProps` as above

src/helpers/getAssetData.ts
```diff
@@ -0,0 +1,41 @@
...
+      assetId,
+      networkUrl,
+    });
+
+    assets[assetId] = normalizeAssetProps({
+      source: data,
+      homeDomain,
+      supportedActions,
+    });
+  }
```
piyalbasu, line 38: ah, for loop because using `forEach` with `await` inside wasn't resolving properly, right? I just had a similar situation. Nothing wrong with your code, but in case you're curious, you can also do this:

```
await Promise.all(
  allAssets.map(async ([assetId, data]) => {
    ...await whatever you need to...
  })
  
)
```

It will wait for each Promise to resolve before moving on to the next
Here's a good explanation on it: https://stackoverflow.com/a/37576787

piyalbasu (approved):

src/components/Balance.tsx
quietbits: I like that, good point. 👍

src/helpers/getAssetData.ts
quietbits: We will keep for loop because we need to make sure the sequence of the assets doesn't change.

PR #106, opened 2021-03-05
SEP-31 Send UX updates

quietbits:


src/components/SendPayment.tsx
```diff
@@ -94,72 +94,70 @@ export const SendPayment = ({
...
+          value={amount}
+          onChange={(e) => setAmount(e.target.value)}
+        />
+        <Input
+          id="send-asset-code"
+          label="Asset code"
+          value={assetCode}
+          onChange={(e) => setAssetCode(e.target.value)}
+        />
+        {asset?.assetType !== "native" && (
```
piyalbasu, line 122: ASSET_TYPE.NATIVE?

piyalbasu (approved):

src/components/SendPayment.tsx
quietbits: 💯 Good catch!

PR #107, opened 2021-03-05
Fix bug when active asset not set after SEP-24 popup closed

quietbits:
The problem was that the active asset was resetting/clearing because the SEP-24 deposit/withdrawal status after closing modal was kept on "SUCCESS" (it should be cleared).
Refactored with better checks and more consistent `useEffect` actions.

### DIFF
--- a/src/components/Assets.tsx
+++ b/src/components/Assets.tsx
@@ -131,9 +131,12 @@ export const Assets = ({
       status: ActionStatus | undefined;
       message: string | React.ReactNode;
     }) => {
+      if (!status) {
+        return;
+      }
+
       if (status === ActionStatus.SUCCESS || status === ActionStatus.ERROR) {
         dispatch(resetActiveAsset());
-        setToastMessage(undefined);
       }
 
       if (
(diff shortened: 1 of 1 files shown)


src/components/Assets.tsx
```diff
@@ -211,12 +218,12 @@ export const Assets = ({
...
   useEffect(() => {
-    if (
-      sep24WithdrawAsset.status === ActionStatus.SUCCESS &&
-      sep24WithdrawAsset.data.currentStatus === "completed"
-    ) {
+    if (sep24WithdrawAsset.status === ActionStatus.SUCCESS) {
       dispatch(resetSep24WithdrawAssetAction());
-      handleRefreshAccount();
+
+      if (sep24WithdrawAsset.data.currentStatus === "completed") {
```
piyalbasu, line 224: `completed` should probably be a const

piyalbasu (approved):

src/components/Assets.tsx
quietbits: Created [an issue](https://github.com/stellar/nebula/issues/134) to do it for all transaction statuses.

PR #114, opened 2021-03-09
Assets update + home domain override

quietbits:
- Add asset modal:
  - accepts home domain and/or issuer public key,
  - added validation (only valid assets will be added through the form).
- Some refactoring to improve add asset flow.
- Disable trust asset action if the account wasn't created.
- Show invalid asset on the UI with option to remove it (previously it was ignored and the only option to remove it was to do it manually from the URL).
  - Adding invalid assets is only possible through URL.
 
image

Home domain override

image


src/components/UntrustedBalance.tsx
```diff
@@ -105,31 +142,57 @@ export const UntrustedBalance = ({
...
+                handleActionChange({
+                  actionId: AssetActionId.TRUST_ASSET,
+                  asset,
+                })
+              }
+              disabled={
+                account.isUnfunded ||
+                Boolean(activeAsset.asset) ||
+                trustAsset.status === ActionStatus.PENDING
+              }
```
piyalbasu, line 167: v small - i liked what you did with `formIsEmpty` above. Since you have 2 similar disabled statements, might be nice to do:
```
const assetNotFound = Boolean(activeAsset.asset) || trustAsset.status === ActionStatus.PENDING; 
// with a more descriptive var name lol

disabled = {account.isUnfunded || assetNotFound}
```

src/components/UntrustedBalance.tsx
```diff
@@ -105,31 +142,57 @@ export const UntrustedBalance = ({
...
+            asset={asset}
+          >
+            <TextButton
+              onClick={() =>
+                handleActionChange({
+                  actionId: AssetActionId.REMOVE_ASSET,
+                  asset,
+                })
+              }
+              disabled={
```
piyalbasu, line 186: and
```
disabled={assetNotFound}
```

piyalbasu (approved):

JakeUrban (approved):
Looks good to me 👍

src/components/UntrustedBalance.tsx
quietbits: Yeah, this one got a little long. 😅 A variable there will make it much more clear. 👍

quietbits:
@piyalbasu @JakeUrban please check this PR again. I know it's a pretty big one but it didn't make sense to split this into multiple PRs because the previous update got changed again, etc. I tried to split the updates in commits, so hopefully, that will help a bit. 🙏

leighmcculloch:
For a developer/user of the demo wallet it would simplify the UI a little if the domain and issuer public key were considered one and the same, and then the user doesn't have to identify that some fields are a one-of required.

Right now the UI looks like this:
Screen Shot 2021-03-18 at 10 50 36 AM

We don't need to explain the OR relationship at the top if you combine the two:
Screen Shot 2021-03-18 at 10 53 07 AM

G addresses and domains are distinguishable, so it should be possible to tell them apart.

src/helpers/getCurrenciesFromDomain.ts
```diff
@@ -0,0 +1,38 @@
+import { Asset, StellarTomlResolver } from "stellar-sdk";
```
piyalbasu, line 1: i didn't know stellar-sdk had a toml resolver! I wrote my own last night because I'm doing something similar in Freighter 😭 😆

src/helpers/getCurrenciesFromDomain.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const getCurrenciesFromDomain = async (
+  homeDomain: string,
```
piyalbasu, line 4: it would actually be great to share this function in Freighter (and other apps that need it). I wonder if it makes sense to add this to stellar-sdk or some other shared util library. I'll create a ticket for myself to start thinking about that. 

I'm sure there are other instances where we're duplicating a lot of the same things

piyalbasu (approved):

src/helpers/getCurrenciesFromDomain.ts
quietbits: Oh no...

src/helpers/getCurrenciesFromDomain.ts
quietbits: 💯 Agree with this.

leighmcculloch: 💯 It would be awesome to add this to the js-stellar-sdk.

PR #115, opened 2021-03-10
Improve error messages

quietbits:
`error.toString()` adds "Error: " prefix to all error messages, `error.message` makes it cleaner.

### DIFF
--- a/src/components/AddAsset.tsx
+++ b/src/components/AddAsset.tsx
@@ -1,6 +1,7 @@
 import { useEffect, useState } from "react";
 import { useHistory } from "react-router-dom";
 import { Button, Heading2, Input, Loader } from "@stellar/design-system";
+import { getErrorMessage } from "helpers/getErrorMessage";
 import { getUntrustedAssetsSearchParam } from "helpers/getUntrustedAssetsSearchParam";
 import { getValidatedUntrustedAsset } from "helpers/getValidatedUntrustedAsset";
 import { log } from "helpers/log";
@@ -56,8 +57,8 @@ export const AddAsset = ({ onClose }: { onClose: () => void }) => {
       );
       setLocalStatus(ActionStatus.SUCCESS);
     } catch (e) {
-      log.error({ title: e.message });
-      setErrorMessage(e.message);
+      log.error({ title: getErrorMessage(e) });
+      setErrorMessage(getErrorMessage(e));
       setLocalStatus(ActionStatus.ERROR);
     }
   };
--- a/src/components/AddAsset.tsx
+++ b/src/components/AddAsset.tsx
@@ -56,8 +56,8 @@ export const AddAsset = ({ onClose }: { onClose: () => void }) => {
       );
       setLocalStatus(ActionStatus.SUCCESS);
     } catch (e) {
-      log.error({ title: e.toString() });
-      setErrorMessage(e.toString());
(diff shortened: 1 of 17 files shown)


src/components/AddAsset.tsx
```diff
@@ -56,8 +56,8 @@ export const AddAsset = ({ onClose }: { onClose: () => void }) => {
       );
       setLocalStatus(ActionStatus.SUCCESS);
     } catch (e) {
-      log.error({ title: e.toString() });
-      setErrorMessage(e.toString());
+      log.error({ title: e.message });
```
quietbits, line 59: @piyalbasu Do you think we should make a helper function to display `error.message` or `error.toString()` if `message` is not available. Errors that we are throwing should have a `message` prop, but there might be cases where it comes from somewhere else and doesn't have it? 🤔

piyalbasu: I think that's a good idea! I'm running into a similar issue in Freighter where `e` doesn't always have a `message` attached to it

src/helpers/getValidatedUntrustedAsset.ts
```diff
@@ -32,8 +33,8 @@ export const getValidatedUntrustedAsset = async ({
 
     asset = `${assetCode}:${homeDomainIssuer}`;
   } catch (e) {
-    log.error({ title: "Issuer domain error: ", body: e.message });
-    throw new Error(e.message);
+    log.error({ title: "Issuer domain error: ", body: getErrorMessage(e) });
+    throw new Error(e);
```
piyalbasu, line 37: Should you be passing Error a string? `throw new Error(getErrorMessage(e))`

Not sure what kind of errors `getIssuerFromDomain` might throw, but you might end up in wrapping something that's already an Error object in another Error object

src/components/AddAsset.tsx
```diff
@@ -56,8 +57,8 @@ export const AddAsset = ({ onClose }: { onClose: () => void }) => {
       );
       setLocalStatus(ActionStatus.SUCCESS);
     } catch (e) {
-      log.error({ title: e.message });
-      setErrorMessage(e.message);
+      log.error({ title: getErrorMessage(e) });
+      setErrorMessage(getErrorMessage(e));
       setLocalStatus(ActionStatus.ERROR);
```
piyalbasu, line 62: very very small, but probably best practice to cache the value of `getErrorMessage(error)` as a const in instances like this and below where the value isn't changing

src/ducks/sep24DepositAsset.ts
```diff
@@ -126,11 +127,11 @@ export const depositAssetAction = createAsyncThunk<
     } catch (error) {
       log.error({
         title: "Deposit failed",
-        body: error.message,
+        body: getErrorMessage(error),
       });
 
       return rejectWithValue({
-        errorString: error.message,
+        errorString: getErrorMessage(error),
```
piyalbasu, line 134: same as above - probably best to cache error when you're logging and then rejecting with the same string

piyalbasu (approved):

src/helpers/getValidatedUntrustedAsset.ts
quietbits: 👍 You're right, I'll fix that. We have more control if we pass along the error string.

PR #116, opened 2021-03-10
Code review fixes 1

quietbits:
- Handle fetching account better (no extra call needed for an unfunded account).
- Remove unused settings: `homeDomain`, `horizonURL`.
- Rename `activeAsset.asset` to `activeAsset.action` for clarity.
- Update SEP-12 origin URL.
- Use stronger types in `ducks/account.ts`.
- Stop polling SEP-24 deposit when status is `pending_external`.

### DIFF
--- a/src/methods/sep24/pollDepositUntilComplete.ts
+++ b/src/methods/sep24/pollDepositUntilComplete.ts
@@ -26,9 +26,11 @@ export const pollDepositUntilComplete = async ({
 
   while (
     !popup.closed &&
-    ![TransactionStatus.COMPLETED, TransactionStatus.ERROR].includes(
-      currentStatus,
-    )
+    ![
+      TransactionStatus.PENDING_EXTERNAL,
+      TransactionStatus.COMPLETED,
+      TransactionStatus.ERROR,
+    ].includes(currentStatus)
   ) {
     // eslint-disable-next-line no-await-in-loop
     const response = await fetch(transactionUrl.toString(), {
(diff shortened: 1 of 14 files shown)


src/methods/sep24/pollDepositUntilComplete.ts
```diff
@@ -112,9 +114,11 @@ export const pollDepositUntilComplete = async ({
...
 
   if (
-    ![TransactionStatus.COMPLETED, TransactionStatus.ERROR].includes(
-      currentStatus,
-    ) &&
+    ![
+      TransactionStatus.PENDING_EXTERNAL,
+      TransactionStatus.COMPLETED,
+      TransactionStatus.ERROR,
+    ].includes(currentStatus) &&
```
piyalbasu, line 121: might be helpful to save 
```
[
      TransactionStatus.PENDING_EXTERNAL,
      TransactionStatus.COMPLETED,
      TransactionStatus.ERROR,
      ]
      ```
as a const

**Review by piyalbasu — APPROVED** [review]

File: src/types/types.d.ts [hunk 2]
```diff
@@ -203,15 +201,15 @@ export interface ClaimableAsset extends Asset {
   claimants: any[];
 }
 
-export interface ActiveAsset {
+export interface ActiveAssetAction {
   id: string;
```
JakeUrban, line 205: Do you think `id` makes sense here, I think `assetCode` might be better, unless you think we could have `id`s that aren't asset codes in the future.

JakeUrban (approved):
Looks good 👏  Just one suggestion

src/types/types.d.ts
quietbits: Yeah, I thought about that and went with a more flexible `id`, which happens to be `assetCode` but it might be something else? 🤔 I'm OK either way. @piyalbasu, how do you feel about it?

piyalbasu: Since it doesn't seem like there are instances of `id` not being an `assetCode`, I like simply calling it `assetCode` now as it's a more descriptive property name.

quietbits: 👍 I'll update, thank you!

PR #124, opened 2021-03-19
Update text + add tooltip to some components

quietbits:
All the components that were brought in (Heading, TextButton, TextLink, etc.) will be moved to the Stellar Design System. It's easier to polish them here and then move to the design system to publish.

### DIFF
--- a/src/components/TextButton/styles.scss
+++ b/src/components/TextButton/styles.scss
@@ -1,0 +1,62 @@
+.TextButtonWrapper {
+  display: flex;
+  align-items: center;
+  position: relative;
+
+  // & > :first-child:not(:last-child) {
+  //   margin-right: 0.3rem;
+  // }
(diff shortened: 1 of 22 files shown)


src/components/TextButton/styles.scss
```diff
@@ -0,0 +1,62 @@
+.TextButtonWrapper {
+  display: flex;
+  align-items: center;
+  position: relative;
+
+  // & > :first-child:not(:last-child) {
+  //   margin-right: 0.3rem;
+  // }
```
piyalbasu, line 8: remove?

piyalbasu (approved):

src/components/TextButton/styles.scss
quietbits: Yes, thank you!

PR #125, opened 2021-03-20
SEP tweaks (home domain, TOML, check info)

quietbits:


src/methods/checkTomlForSep.ts
```diff
@@ -0,0 +1,62 @@
+import { log } from "helpers/log";
+import { getHomeDomainFromAssetIssuer } from "methods/getHomeDomainFromAssetIssuer";
+import { getToml } from "methods/getToml";
+import { TomlFields, AnyObject } from "types/types.d";
+
+export const checkTomlForSep = async ({
```
JakeUrban, line 6: Maybe `checkTomlForFields` is a better name?

src/methods/checkTomlForSep.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (!homeDomainParam) {
+    homeDomainParam = await getHomeDomainFromAssetIssuer({
+      assetIssuer,
+      networkUrl,
+    });
+  }
+
+  if (!homeDomainParam) {
+    throw new Error("Home domain was not found and is required");
+  }
```
JakeUrban, line 30: I think a more appropriate pattern would be to raise an error from `getHomeDomainFromAssetIssuer()` instead of checking if `homeDomainParam` is defined after calling a function that looks like it should always return one.

This also matches the pattern used in `getToml()`

src/methods/sep31Send/checkInfo.ts
```diff
@@ -37,10 +37,6 @@ export const checkInfo = async ({
     throw new Error("No `fields` object specified in /info");
   }
 
-  if (!asset.fields.transaction || isEmptyObject(asset.fields.transaction)) {
-    throw new Error("No transaction fields specified");
-  }
```
JakeUrban, line 42: All SEP-31 deployments must have the `fields` key and `transaction` subkey, but it's ok if the `transaction` object is empty.

JakeUrban (commented):
Some minor suggestions but otherwise LGTM

src/ducks/sep24WithdrawAsset.ts
```diff
@@ -11,48 +11,70 @@ import {
...
+        toml: tomlResponse,
+        assetCode,
+      });
 
       // SEP-10 start
       const challengeTransaction = await sep10AuthStart({
         authEndpoint: tomlResponse.WEB_AUTH_ENDPOINT,
-        secretKey,
+        publicKey,
+        homeDomain,
```
piyalbasu, line 77: Doesn't necessarily have to be in this PR, but there's a fair amount of repetition between the withdraw and deposit flow. Might be nice to wrap all the checking and signing stuff into a helper function

piyalbasu (approved):

JakeUrban (approved):
LGTM!

PR #126, opened 2021-03-22
Configuration modal + claimableBalanceSupported search param

quietbits:
image


src/methods/sep24/interactiveDepositFlow.ts
```diff
@@ -1,23 +1,27 @@
...
-}) => {
+  claimableBalanceSupported,
+}: InteractiveDepositFlowProps) => {
   const formData = new FormData();
   const postDepositParams = {
     asset_code: assetCode,
     account: publicKey,
     lang: "en",
-    claimable_balance_supported: "true",
+    claimable_balance_supported: claimableBalanceSupported.toString(),
```
quietbits, line 24: Do we need to handle cases where this is `false`?

piyalbasu (approved):

PR #133, opened 2021-04-13
on re-render, set value of dropdown to "", or initial

piyalbasu:
https://user-images.githubusercontent.com/6789586/114484752-c5e08900-9bd8-11eb-9b49-5518989012bf.mov

### DIFF
--- a/src/components/BalanceRow.tsx
+++ b/src/components/BalanceRow.tsx
@@ -81,6 +81,7 @@ export const BalanceRow = ({
               id={`${assetString}-actions`}
               onChange={(e) => onAction(e.target.value, asset)}
               disabled={disabled}
+              value=""
             >
               <option value="">Select action</option>
               {!isUntrusted && (
--- a/src/components/BalanceRow.tsx
+++ b/src/components/BalanceRow.tsx
@@ -1,4 +1,4 @@
-import { ReactNode } from "react";
+import React, { ReactNode, useEffect, useState } from "react";
 import { Select } from "@stellar/design-system";
 import { TextLink } from "components/TextLink";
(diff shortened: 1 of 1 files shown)


src/components/BalanceRow.tsx
```diff
@@ -81,6 +81,7 @@ export const BalanceRow = ({
               id={`${assetString}-actions`}
               onChange={(e) => onAction(e.target.value, asset)}
               disabled={disabled}
+              value=""
```
piyalbasu, line 84: I acknowledge this is very simplistic :D , but if we only care about this dropdown defaulting to the initial value on re-render, this will do the trick. 

I *could* save the dropdown's state and then update after `onAssetAction`'s completion, but that seems like overkill as this appears to be work for all the use cases I tried. Am I missing a scenario where we would need the added flexibility of saving in state?

quietbits: We might want to check with the Product if this is acceptable because in this case user will never see the action selected (re-render happens when we show the modal).

image

quietbits: 🤔 Maybe we could reset the value only when it is not `isActive`.

piyalbasu: Ah good catch - I knew the dropdown was resetting when we show the modal, but on my screen the modal was always covering the dropdown. I'll put in a better fix!

shanzzam: That is the desired behavior. Thanks for pointing that out @quietbits.

piyalbasu:
Updated to maintain dropdown value until modal is dismissed
Screen Shot 2021-04-13 at 12 56 11 PM

quietbits (approved):
🎉

PR #134, opened 2021-04-15
add standard SDF metrics handling and landing page tracking

piyalbasu:


src/constants/metricNames.ts
```diff
@@ -0,0 +1,3 @@
+export const METRIC_NAMES = {
+  viewHome: "loaded screen: home",
```
piyalbasu, line 2: @shanzzam Let me know if you would like to change this label! This is the event name that will show up in Amplitude

src/helpers/metrics.ts
```diff
@@ -0,0 +1,64 @@
+import throttle from "lodash/throttle";
```
piyalbasu, line 1: this is lifted from Freighter/Laboratory, which each have almost this exact code. I think it's best to keep this project in parity with our other tools.

@quietbits this is a prime example of something that can go into the shared "helpers" repo we've been talking about :D

src/pages/Landing.tsx
```diff
@@ -26,6 +28,10 @@ export const Landing = () => {
   const location = useLocation();
   const history = useHistory();
 
+  useEffect(() => {
+    emitMetric(METRIC_NAMES.viewHome);
+  }, []);
+
```
piyalbasu, line 34: Carl created a custom middleware in Freighter that automatically emits an event on routing. Might be nice to do something similar in a future iteration

src/constants/metricNames.ts
shanzzam: This is fine. I guess as there's only one page we're ever tracking this works! We can add metrics for how the rest of the content is accessed via keypair selection later.

quietbits (approved):

PR #135, opened 2021-04-15
init Sentry and capture error strings when logged

piyalbasu:
We can continue to tune our level of error capturing in the future releases. This seems solid for a v1

### DIFF
--- a/src/ducks/sendPayment.ts
+++ b/src/ducks/sendPayment.ts
@@ -1,9 +1,9 @@
 import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
 import { Horizon } from "stellar-sdk";
-import { captureException as sentryCaptureException } from "@sentry/browser";
 import { RootState } from "config/store";
 import { settingsSelector } from "ducks/settings";
 import { getErrorString } from "helpers/getErrorString";
+import { log } from "helpers/log";
 import { submitPaymentTransaction } from "methods/submitPaymentTransaction";
 import {
   ActionStatus,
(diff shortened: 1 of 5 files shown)


src/helpers/log.ts
```diff
@@ -59,6 +60,7 @@ export const log = {
   },
 
   error: ({ title, body = "" }: { title: string; body?: string | object }) => {
+    sentryCaptureMessage(title);
```
piyalbasu, line 63: @quietbits has done a great hob making sure thrown errors eventually bubble up to this logging method, so let's just capture everything here

src/ducks/sendPayment.ts
```diff
@@ -28,6 +29,7 @@ export const sendPaymentAction = createAsyncThunk<
         isPubnet: pubnet,
       });
     } catch (error) {
+      sentryCaptureException(error);
```
piyalbasu, line 32: this is the only spot I found where an error would not be caught by the logger

quietbits: I think it might be because we are logging the error on the UI side. 🤔 Can't think of any other reason.

quietbits (approved):
🎉

src/ducks/sendPayment.ts
quietbits: Actually, I just tested it and the error wasn't logged. @piyalbasu Could you please add it here? Then we can remove the Sentry from here. It was an oversight on my part for missing it here. 😬

PR #138, opened 2021-04-21
Support localhost home domain override

quietbits:


src/helpers/searchKeyPairStringToArray.ts
```diff
@@ -16,7 +17,7 @@ export const searchKeyPairStringToArray = (
     const values = paramArr
       .splice(1, paramArr.length - 1)
       .reduce((paramRes, val) => {
-        const [key, value] = val.split(":");
+        const [key, value] = val.split(URL_KEY_PAIR_SEPARATOR_CHAR);
```
JakeUrban, line 20: `URL_KEY_PAIR_SEPARATOR_CHAR` is set to `>`, but the previously used value is `:`.

I think `URL_KEY_PAIR_SEPARATOR_CHAR` should be `:` right? I'm looking at the example in the comment on line 4. If not lets update the example.

src/helpers/searchParam.ts
```diff
@@ -107,7 +111,9 @@ const updateKeyPair = ({
     const updatedValuesString = updatedValuesArray.reduce(
       (result: string[], asset) => [
         ...result,
-        `${asset.assetString}:${getKeyPairString(asset)}`,
+        `${asset.assetString}${URL_ITEM_SEPARATOR_CHAR}${getKeyPairString(
```
JakeUrban, line 114: If `asset.assetString` is `<code>:<issuer pub key>`, what does `getKeyPairString(asset)` return, the home domain key and value?

If `asset.assetString` is just `<code>`, then shouldn't we use `URL_KEY_PAIR_SEPARATOR_CHAR`?

src/helpers/normalizeHomeDomainUrl.ts
```diff
@@ -0,0 +1,16 @@
...
+  // default localhost to http instead of https
+  if (_homeDomain.includes("localhost")) {
+    _homeDomain = _homeDomain.startsWith("http")
+      ? _homeDomain
+      : `http://${_homeDomain}`;
+  } else {
+    _homeDomain = _homeDomain.startsWith("http")
+      ? _homeDomain
+      : `https://${_homeDomain}`;
+  }
```
piyalbasu, line 13: I spent a few min coming up with a way to do this in fewer lines, but I honestly think keeping it the way you wrote it is more clear haha

JakeUrban (approved):
One comment and one question. I tested it out and everything works. If the question is resolved then it LGTM 👍

src/helpers/normalizeHomeDomainUrl.ts
piyalbasu: This was the shorter version I came up with, which is actually kind of confusing:
```  
if (!_homeDomain.startsWith("http")) {
    _homeDomain = _homeDomain.includes("localhost") ? `http://${_homeDomain}` : `https://${_homeDomain}` 
  }
```

piyalbasu (approved):

src/helpers/searchKeyPairStringToArray.ts
quietbits: Thanks, will update the comment!

src/helpers/normalizeHomeDomainUrl.ts
quietbits: It's funny, I did exactly the same thing and went with the more clear solution. With simple `if else` there is no question about what is going on.
There are two things ( `localhost` and `.startsWith("http")`) we need to check because provided home domain could have `http(s)` in which case we leave it as is.

src/helpers/searchParam.ts
quietbits: This was actually a bug the way we had it before. The item format is `assetString|key>value`. I'll add a comment with the result we're getting for clarity.

We're building query params back, so the format will be `assetSrting|key>value`, `getKeyPairString(asset)` returns `key>value|key1>value1` pairs string.

JakeUrban (approved):

src/helpers/searchKeyPairStringToArray.ts
shanzzam: @JakeUrban or @quietbits what are the steps to test this locally? I know I need a local instance running but is there a tutorial somewhere that can help me do that? I'm curious to test this although I KNOW @JakeUrban has given it the 💯 approval!

JakeUrban: You would need to have an anchor running locally, its probably easier for one of us to demo this to you.

PR #140, opened 2021-04-22
Adding testnet usage instructions

lydiat:
Rewriting intro to acknowledge launch, adding instructions on how to set up a testnet account

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ You can use the demo wallet to test Hosted Deposit and Withdrawal (SEP-24) and C
 ### Demo-ing Cross-Border Payments (SEP-31) on Testnet
 1. Follow the steps above in order to establish an amount of SRT to send.
 2. Select “SEP-31 Send” from the dropdown for your SRT asset and click "Start" in the modal.
-3. Enter the requested information (none of it has to be real, this is only to show the fields that the sending anchor requires).
+3. Enter the requested information in the pop-up - none of the info has to be real for this testanchor.stellar.org demo, this is only to show the fields required. When testing another anchor you may need to adhere to their validation requirements.
 4. If the payment has been successfully sent you'll see "SEP-31 send payment completed" in the logs.
 
 ## Getting Started Building A Wallet From This Demo
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Stellar Demo Wallet
 
-This Stellar Demo Wallet app will soon be the defacto application to use when
-testing anchor services interactively. If you would live to automate testing of
+This Stellar Demo Wallet is our newly rebuilt application for
+interactively testing anchor services. If you would like to automate testing of
 your anchor service, check out the SDF's
 [anchor validation suite](https://github.com/stellar/transfer-server-validator)
 viewable at [anchor-validator.stellar.org](anchor-validator.stellar.org).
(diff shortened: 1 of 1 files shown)


JakeUrban (changes requested):
There are detailed instructions for the `testanchor.stellar.org` SEP-24 flow but not for SEP-31. Can we add that?

quietbits (approved):

lydiat:
> There are detailed instructions for the `testanchor.stellar.org` SEP-24 flow but not for SEP-31. Can we add that?

SEP-31 is right below SEP-24, but is much smaller because it's "do the above stuff and then"

JakeUrban:
@lydiat SEP-31 instructions stop at selecting "Start", but SEP-24's instructions go into detail about how to complete the flow for a transaction with SRT. Can we do the same for SEP-31? Inputting send/receiver information etc.?

README.md
```diff
@@ -10,10 +10,29 @@ This project was originally created for the
...
+3. Click “Add trustline” - this will allow you to hold SRT.
+4. Select “SEP-24 deposit” from the dropdown for your SRT asset and click "Start" in the modal.
+5. You’ll be asked for some KYC info during this - it doesn’t have to be real, but the interface will want a valid email.
+6. Click “Skip confirmation” - skipping it won't be possible in live integrations but helps the process move ahead in the demo.
+7. Enter a number into the amount and click "Submit". 
+8. Leave the window open while you wait to see the deposit of SRT go through - you can close when you see “Status” is complete and you have SRT.
+
+### Demo-ing Cross-Border Payments (SEP-31) on Testnet
+1. Follow the steps above in order to establish an amount of SRT to send.
+2. Select “SEP-31 Send” from the dropdown for your SRT asset and click "Start" in the modal.
```
shanzzam, line 33: May want to note that info in the "Sender and receiver info" modal isn't real, similar to Step 5 in the SEP-24 deposit/withdraw steps above

shanzzam (commented):
Just one note on sender/receiver modal info

lydiat:
Ahh, good point - I'll flesh out the SEP-31 part 👀

README.md
```diff
@@ -10,10 +10,32 @@ This project was originally created for the
...
+5. If your browser doesn't already, make sure it allows pop-ups - this is how the demo wallet requests KYC info. 
+6. Enter your name and email in the pop-up - this information doesn't need to be real, but the interface will want a valid email.
+7. Click “Skip confirmation” - skipping it won't be possible in live integrations but helps the process move ahead in the demo.
+8. Enter a number into the amount and click "Submit". 
+9. Leave the pop-up window open while you wait to see the deposit of SRT go through - you can close when you see “Status” is complete and you have SRT.
+
+### Demo-ing Cross-Border Payments (SEP-31) on Testnet
+1. Follow the steps above in order to establish an amount of SRT to send.
+2. Select “SEP-31 Send” from the dropdown for your SRT asset and click "Start" in the modal.
+3. Enter the requested information (none of it has to be real, this is only to show the fields that the sending anchor requires).
```
JakeUrban, line 35: It doesn't have to be real for our anchor, but it could need to be real data for another anchor. I think we should make this distinction.

JakeUrban (changes requested):

README.md
lydiat: As I was testing I noticed you don't have to enter ANYTHING for testanchor.stellar.org, although you'll get in trouble once it's run without an amount. We should probably at least require those fields to be filled, and validate the email for parity with SEP-24.

JakeUrban (approved):

PR #147, opened 2021-04-28
SEP-31 support selecting customer types

quietbits:
- Updated `sep31Send` flow to add an extra step for selecting customer type (if there are multiple options available).
- Added `CAN_PROCEED` status to `ActionStatus`, which is similar to `SUCCESS` but happens during the process (`SUCCESS` should happen only once at the end of the process).
- Added `RadioButton` component.
- Added a modal to select sender/receiver types.

image

### DIFF
--- a/src/components/RadioButton/index.tsx
+++ b/src/components/RadioButton/index.tsx
@@ -1,0 +1,14 @@
+import React from "react";
+import "./styles.scss";
+
+interface RadioButtonProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  id: string;
+  label: string;
+}
+
+export const RadioButton = ({ id, label, ...props }: RadioButtonProps) => (
+  <div className="RadioButton">
+    <input type="radio" id={id} {...props} />
+    <label htmlFor={id}>{label}</label>
+  </div>
+);
--- a/src/components/RadioButton/index.tsx
+++ b/src/components/RadioButton/index.tsx
@@ -3,7 +3,7 @@ import "./styles.scss";
 
 interface RadioButtonProps extends React.InputHTMLAttributes<HTMLInputElement> {
   id: string;
-  label: string;
+  label: string | React.ReactNode;
 }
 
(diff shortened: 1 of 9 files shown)


src/components/RadioButton/index.tsx
```diff
@@ -0,0 +1,14 @@
...
+import "./styles.scss";
+
+interface RadioButtonProps extends React.InputHTMLAttributes<HTMLInputElement> {
+  id: string;
+  label: string | React.ReactNode;
+}
+
+export const RadioButton = ({ id, label, ...props }: RadioButtonProps) => (
+  <div className="RadioButton">
+    <input type="radio" id={id} {...props} />
```
JakeUrban, line 11: ❓ Will `{...props}` render with the correct HTML tag syntax?

src/components/Sep31Send.tsx
```diff
@@ -49,76 +84,185 @@ export const Sep31Send = () => {
...
+    event: React.MouseEvent<HTMLButtonElement, MouseEvent>,
+  ) => {
+    setErrorMessage("");
+    const { sender, receiver } = customerTypes;
+    event.preventDefault();
+
+    if (!sender || !receiver) {
+      setErrorMessage("Sender and receiver types are required");
+      return;
+    }
```
JakeUrban, line 113: A receiving anchor could only require KYC for one of the sending & receiving customers, or neither.

JakeUrban (changes requested):
I only found one issue, which is that we need to adjust the UI depending on whether or not the receiving anchor requires KYC for the sender, receiver, or both.

src/components/RadioButton/index.tsx
quietbits: Yes, all the native input props will render correctly (name, value, etc).

src/components/Sep31Send.tsx
quietbits: I'll add better validation 👍

JakeUrban (approved):
LGTM

PR #151, opened 2021-05-10
Identify SEP-8 regulated assets

marcelosalloum:
### What

Find out if the given asset is regulated and sign that accordingly in the UI.

### Screenshot

Screen Shot 2021-05-10 at 13 22 06

### DIFF
--- a/src/helpers/getAssetSettingsFromToml.ts
+++ b/src/helpers/getAssetSettingsFromToml.ts
@@ -39,6 +39,7 @@ export const getAssetSettingsFromToml = async ({
   if (homeDomain) {
     const toml = await getToml(homeDomain);
     const {
+      CURRENCIES,
       TRANSFER_SERVER,
       TRANSFER_SERVER_SEP0024,
       DIRECT_PAYMENT_SERVER,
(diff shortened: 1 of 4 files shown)


Zagan202 (approved):
LGTM

src/helpers/getAssetSettingsFromToml.ts
```diff
@@ -56,3 +58,22 @@ export const getAssetSettingsFromToml = async ({
     ...{ supportedActions },
   };
 };
+
+const isSEP8Asset = ({
```
quietbits, line 62: Super nitpicky, we have a convention of using "pure" camel case convention (not having SEP in all caps, for example), so this function would be `isSep8Asset`. 😊

quietbits (approved):

src/helpers/getAssetSettingsFromToml.ts
marcelosalloum: SGTM! Updated that in https://github.com/stellar/stellar-demo-wallet/pull/151/commits/c016d43e5fe254cb10c46e90a5e323555c3f33dc.

PR #162, opened 2021-05-19
GitHub link in footer

quietbits:
- Add GitHub link to the the footer
- Remove announcement banner

### DIFF
--- a/src/components/Footer.tsx
+++ b/src/components/Footer.tsx
@@ -34,6 +34,16 @@ export const Footer = () => {
             >
               Privacy Policy
             </TextLink>
+
+            <TextLink
+              variant={TextLinkVariant.secondary}
+              href="https://github.com/stellar/stellar-demo-wallet"
+              rel="noreferrer"
+              target="_blank"
+              title="Check out our GitHub repo for more information and to log issues"
+            >
+              GitHub
+            </TextLink>
           </div>
 
           {account.isAuthenticated && (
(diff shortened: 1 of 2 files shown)


src/components/Footer.tsx
```diff
@@ -34,6 +34,16 @@ export const Footer = () => {
             >
               Privacy Policy
             </TextLink>
+
+            <TextLink
+              variant={TextLinkVariant.secondary}
+              href="https://github.com/stellar/stellar-demo-wallet"
+              rel="noreferrer"
+              target="_blank"
+              title="Check out our GitHub repo for more information and to log issues"
```
leighmcculloch, line 43: I think this should be `the`, rather than `our`. Feels too personal.
```suggestion
              title="Check out the GitHub repo for more information and to log issues"
```

What 'more information' are we hoping to give to folks? Who is the audience of this call to action?

I defer to whoever is deciding about copy. 😄

leighmcculloch (approved):
The call to action in the footer feels a little non-specific to who the audience is, and what we're hoping they will want to discover, see note inline, but regardless 👍🏻.

src/components/Footer.tsx
shanzzam: @leighmcculloch the main audience to date have been anchor developers testing implementation against a mock client (this). That's still the aim as well as extending this to a learning tool. We don't want the full banner at the top of the page but we do want users to know they can log bugs in this repo. A README is here as well although we need to float instructions on the main page as well.

piyalbasu (approved):

piyalbasu:
@quietbits Is the plan to release this before `release/1.2.0`? If this can wait, we should merge this into the `release/1.2.0` branch

quietbits:
@shanzzam 👆  Do we want to release it before 1.2.0?

marcelosalloum:
> @shanzzam 👆 Do we want to release it before 1.2.0?

I'd prefer to use our new release strategy and merge these changes into `release/1.2.0`.

PR #163, opened 2021-05-20
Feature/sep 6 deposit

piyalbasu:
https://user-images.githubusercontent.com/6789586/119023973-9576d200-b970-11eb-955b-7fa633c8f30b.mov


piyalbasu:
I will not be merging this until we develop a release strategy for v1.2.0

src/components/Balance.tsx
```diff
@@ -114,6 +119,22 @@ export const Balance = ({
...
+        };
+        break;
+      case AssetActionId.SEP6_WITHDRAW:
+        props = {
+          ...defaultProps,
+          title: `SEP-6 withdrawal ${balance.assetCode}`,
+          description: `Start SEP-6 withdrawal of ${balance.assetCode}?`,
+          callback: () => handleSep24Withdraw(balance),
+        };
+        break;
```
piyalbasu, line 137: this PR covers SEP-6 deposit of a trusted asset. A follow up PR will cover untrusted assets using claimable balances

src/components/Sep6Send.tsx
```diff
@@ -0,0 +1,183 @@
...
+    depositType: {},
+    fields: {},
+  });
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (sep6DepositAsset.status === ActionStatus.NEEDS_INPUT) {
+      setFormData({
+        depositType: {
+          type: sep6DepositAsset.data.depositTypes.type.choices[0],
```
piyalbasu, line 30: default formData to the first `choice` as some assets (like SRT) will only have one

src/ducks/sep6DepositAsset.ts
```diff
@@ -0,0 +1,313 @@
...
+  ActionStatus,
+  Sep6DepositAssetInitialState,
+  Sep6DepositResponse,
+  RejectMessage,
+  TomlFields,
+  CheckInfoType,
+  AnyObject,
+} from "types/types.d";
+
+export const initiateSendAction = createAsyncThunk<
```
piyalbasu, line 28: this is following the general pattern of SEP-24

src/ducks/sep6DepositAsset.ts
```diff
@@ -0,0 +1,313 @@
...
+          ...payload,
+          kycServer: webAuthTomlResponse.KYC_SERVER,
+          token,
+        };
+
+        if (sep12Fields) {
+          payload = {
+            ...payload,
+            fields: { ...payload.fields, ...sep12Fields },
+          };
```
piyalbasu, line 145: add any required fields found in `/customer` endpoint

src/ducks/sep6DepositAsset.ts
```diff
@@ -0,0 +1,313 @@
...
+  { rejectValue: RejectMessage; state: RootState }
+>(
+  "sep6DepositAsset/submitSep6DepositFields",
+  async ({ depositType, fields }, { rejectWithValue, getState }) => {
+    try {
+      const { secretKey } = accountSelector(getState());
+      const { data } = sep6DepositSelector(getState());
+      const { kycServer, token } = data;
+
+      if (Object.keys(fields).length) {
```
piyalbasu, line 180: if we have any customer info fields, submit them. Otherwise, just record the deposit type that the user selected

src/methods/checkDepositWithdrawInfo.ts
```diff
@@ -0,0 +1,31 @@
+import { get } from "lodash";
```
piyalbasu, line 1: I didn't touch much code outside of SEP-6 flow, but I did combine the `checkInfo` methods for SEP-6 and SEP-24 as they are doing the same thing and expect the same payload

src/methods/sep12/collectSep12Fields.ts
```diff
@@ -0,0 +1,54 @@
+import { log } from "helpers/log";
```
piyalbasu, line 1: I also moved the SEP-12 methods into their own folder so SEP-6 and SEP-31 could both use them

src/methods/sep12/collectSep12Fields.ts
```diff
@@ -0,0 +1,54 @@
...
+    headers: {
+      Authorization: `Bearer ${token}`,
+      Origin: "https://demo-wallet.stellar.org",
+    },
+  });
+  const resultJson = await result.json();
+
+  log.response({ title: "GET `/customer`", body: resultJson });
+
+  if (isNewCustomer && resultJson.status !== "NEEDS_INFO") {
```
piyalbasu, line 40: one small change to this code: I added this `isNewCustomer` boolean as SEP-31 wants this method to throw error if it doesn't need more info. SEP-6, however, should continue working if no add'l info is needed

src/components/Sep6Send.tsx
```diff
@@ -0,0 +1,183 @@
...
+  sep6DepositAction,
+} from "ducks/sep6DepositAsset";
+import { useRedux } from "hooks/useRedux";
+import { ActionStatus } from "types/types.d";
+
+export const Sep6Send = () => {
+  const { sep6DepositAsset } = useRedux("sep6DepositAsset");
+  const { depositResponse } = sep6DepositAsset;
+
+  const [formData, setFormData] = useState<any>({
```
quietbits, line 20: Maybe we could be more specific about the type here.

src/components/Sep6Send.tsx
```diff
@@ -0,0 +1,183 @@
...
+          }
+        >
+          SEP-6 Required Info
+        </Heading2>
+        <div className="ModalBody">
+          {Object.entries(sep6DepositAsset.data.depositTypes || {}).map(
+            ([id, input]) => (
+              <>
+                <label>{input.description}</label>
+                <Select id={id} key={id} onChange={handleDepositTypeChange}>
```
quietbits, line 120: `<Select>` has a `label` prop, can't we use it here?

src/ducks/sep6DepositAsset.ts
quietbits: Are we doing anything with submitted fields or the response?

src/ducks/sep6DepositAsset.ts
```diff
@@ -0,0 +1,311 @@
...
+  Sep6DepositAssetInitialState,
+  Sep6DepositResponse,
+  RejectMessage,
+  TomlFields,
+  CheckInfoType,
+  AnyObject,
+} from "types/types.d";
+
+export const initiateSendAction = createAsyncThunk<
+  { fields: {}; status: ActionStatus },
```
quietbits, line 29: We could use `AnyObject` type for `fields`.

quietbits (approved):
Looks great, just left a few minor comments/questions. 🙌

src/ducks/sep6DepositAsset.ts
piyalbasu: We're not - it looks like we just care about it not throwing an error. This request does respond with a generated `id`, but we don't appear to have a use for it, nor the fields that we submitted, as long as the response is a 200

src/components/Sep6Send.tsx
piyalbasu: Ah, missed that! Yes, I should use the label prop

PR #175, opened 2021-05-26
Support all SEP-9 fields when handling SEP-8 "action_required" response

marcelosalloum:
### What

Support all SEP-9 fields when handling SEP-8 "action_required" response.

### Why

In order to have a complete implementation of SEP-8 we should be able to send binary and date fields as well.


piyalbasu (approved):

src/helpers/Sep9Fields.ts
```diff
@@ -0,0 +1,214 @@
...
+    name: "id_type",
+    type: Sep9FieldType.STRING,
+    description: "passport, drivers_license, id_card, etc...",
+  },
+  {
+    name: "id_country_code",
+    type: Sep9FieldType.COUNTRY_CODE,
+    description:
+      "country issuing passport or photo ID as ISO 3166-1 alpha-3 code",
+  },
```
quietbits, line 159: NIT: was it intentional that these two descriptions start with a lowercase letter?

quietbits (approved):

src/helpers/Sep9Fields.ts
marcelosalloum: Good catch! I copied/pasted it from SEP-9 and forgot to adjust these descriptions.
Fixed in https://github.com/stellar/stellar-demo-wallet/pull/175/commits/6c52210215d87aa466765df092ef4866e5c5f899.

PR #185, opened 2021-06-08
Update Sentry tracingOrigins

quietbits:
Take 2 fixing this (empty array in the previous fix was ignored). 🤞


JakeUrban:
What does this regular express match?

piyalbasu (approved):

quietbits:
@JakeUrban the regex expression is "match `/` that is not followed by another `/`". It feels like it should be pretty safe as it should be unlikely to get a link like that.
1. `//stellar.org/test` - no match
2. `/stellar.org/test` - match

JakeUrban (approved):

PR #189, opened 2021-06-18
SEP-6 deposit updates

quietbits:
- Add amount input
- Show min/max amount
- Button to trigger polling action (it was automatic before)

### DIFF
--- a/src/components/Sep6/Sep6Deposit.tsx
+++ b/src/components/Sep6/Sep6Deposit.tsx
@@ -63,12 +63,6 @@ export const Sep6Deposit = () => {
     }
   }, [sep6DepositAsset.status, depositTypeChoices, dispatch]);
 
-  useEffect(() => {
-    if (sep6DepositAsset.status === ActionStatus.CAN_PROCEED) {
-      dispatch(sep6DepositAction());
-    }
-  }, [sep6DepositAsset.status, dispatch]);
-
   const resetLocalState = () => {
     setFormData(formInitialState);
   };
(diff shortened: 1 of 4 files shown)


shanzzam:
Testing this: I'm getting the KYC collection popup every time. Aren't I supposed to get this only the first time?

quietbits:
I'm getting KYC every time now as well, I'm pretty sure it wasn't happening last week. 🤔 I'm seeing the same on the release PR preview. @JakeUrban has anything changed on the backend?

JakeUrban:
I recently added support for [SEP-12 property verification](https://github.com/stellar/stellar-protocol/commit/2d46fe87fa9aa512610078eda51dc52573ed754b#), here's the [Polaris commit](https://github.com/stellar/django-polaris/commit/448e012520895e1d81b31da634c103eb83509d9d).

How does the demo wallet determine which fields need to be collected?

quietbits:
@JakeUrban We're checking if there are any `fields` returned for `/customer`. Should we be checking the status of each field returned? Before this update, the fields were not returned at all if they didn't need to be submitted.

@shanzzam Should we create a separate issue for this?

JakeUrban:
I'm realizing there was some ambiguity introduced with the SEP-12 change, so I'm making an issuing on the stellar-protocol repo to discuss potential clarifications.

For now, the demo wallet should first check if the customer's status is `NEEDS_INFO`. If it is, then information needs to be collected, otherwise the anchor is either processing info or the customer was rejected. 

When the customer's status is `NEEDS_INFO`, we should check for the presence of field-level `status` values. `NOT_PROVIDED` fields should be collected in the modal, and fields without `status` values can be assumed to be `NOT_PROVIDED`.

shanzzam:
This looks good to me!

piyalbasu (approved):

PR #193, opened 2021-06-23
SEP-12: Check customer fields status

quietbits:


src/methods/sep12/collectSep12Fields.ts
```diff
@@ -43,12 +44,42 @@ export const collectSep12Fields = async ({
...
 
+  const fieldsToCollect = Object.entries(resultJson.fields ?? {}).reduce(
+    (collectResult: any, field: any) => {
+      const [key, props] = field;
+
+      if (
+        !props.status ||
+        [
+          Sep12CustomerFieldStatus.NOT_PROVIDED,
+          Sep12CustomerFieldStatus.REJECTED,
```
quietbits, line 55: Do we want to include rejected fields as well?

JakeUrban: If the customer's status is `REJECTED`, then that is final (the user will never pass). If the customer's status is `NEEDS_INFO`, then individual fields that are `REJECTED` can be re-collected.

It doesn't look like we're checking the customer status other than when the customer is new, so we'll need to adjust the logic here.

JakeUrban (changes requested):
Looks good other than the topic of REJECTED fields you asked about.

Also, this is unrelated by I see on line 24 that we may be sending an `undefined` memo. If `memo` is not defined, we shouldn't be sending it in the request.

quietbits:
@JakeUrban Thanks for your comments! The memo looks good, we're not submitting it if it's not provided (spreading `{}` won't produce anything in this case). It was good to double check. 👍

src/methods/sep12/collectSep12Fields.ts
```diff
@@ -37,18 +38,47 @@ export const collectSep12Fields = async ({
...
   if (resultJson.fields) {
     log.instruction({
-      title: "Received the following fields",
+      title: "Received the following customer fields",
       body: resultJson.fields,
     });
   }
 
-  return resultJson.fields;
+  if (Object.keys(fieldsToCollect).length > 0) {
```
piyalbasu, line 72: sooooo small - but this could be simplified to `(Object.keys(fieldsToCollect).length )`

piyalbasu (approved):

JakeUrban (approved):
We will need to make some adjustments to this logic in the near future once we make changes to the SEP-12 `GET /customer` response format but this looks good for what SEP-12 is now 👍

PR #194, opened 2021-06-24
Do not show actions for untrusted regulated assets

quietbits:
image

### DIFF
--- a/src/components/BalanceRow.tsx
+++ b/src/components/BalanceRow.tsx
@@ -54,6 +54,82 @@ export const BalanceRow = ({
     }
   };
 
+  const renderActionsSelect = () => {
+    if (onAction) {
+      // Regulated asset needs to be trusted first
+      if (isUntrusted && asset.supportedActions?.sep8) {
+        return null;
+      }
+
+      return (
+        <div className="BalanceCellSelect">
+          <Select
+            id={`${assetString}-actions`}
+            onChange={handleSelectChange}
+            disabled={disabled}
+            value={selectValue}
+          >
+            <option value="">Select action</option>
+            {!isUntrusted && !asset.supportedActions?.sep8 && (
+              <option value={AssetActionId.SEND_PAYMENT}>Send payment</option>
+            )}
+
+            {supportedActions?.sep6 && (
+              <>
+                <option value={AssetActionId.SEP6_DEPOSIT}>
+                  SEP-6 Deposit
+                </option>
+                {!isUntrusted && (
+                  <option value={AssetActionId.SEP6_WITHDRAW}>
+                    SEP-6 Withdraw
+                  </option>
+                )}
+              </>
+            )}
+
+            {asset.supportedActions?.sep8 && (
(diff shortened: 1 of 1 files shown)


src/components/BalanceRow.tsx
```diff
@@ -54,6 +54,82 @@ export const BalanceRow = ({
     }
   };
 
+  const renderActionsSelect = () => {
+    if (onAction) {
+      // Regulated asset needs to be trusted first
+      if (isUntrusted && asset.supportedActions?.sep8) {
+        return null;
+      }
```
marcelosalloum, line 62: Do we want the SEPs to be available before adding a trustline to a non-regulated asset?

I'm wondering if this change is desired only for regulated assets or if for all assets.

marcelosalloum (commented):
PR looks good! I just have a question below

src/components/BalanceRow.tsx
quietbits: There are some SEPs that can work without a trustline (it is created during deposit, for example). If SEP-8 is not supported for that asset, we would show everything as it was before.

piyalbasu (approved):

src/components/BalanceRow.tsx
marcelosalloum: Ok, got it. Now I can see SEP-6 and SEP-24 actions are supported when there's no trustline: https://stellar-demo-wallet-pr194.previews.kube001.services.stellar-ops.com/account?secretKey=SCIAIFDX7QYORHOWZNNU24DXQXTNP2AABHGNRJ5WPZ6AC6VXXR3VCTHT&untrustedAssets=MULT%3AGDLD3SOLYJTBEAK5IU4LDS44UMBND262IXPJB3LDHXOZ3S2QQRD5FSMM

I think this behaviour should remain the same across SEP-8 and non-SEP-8 assets. There could be an anchor implementing SEP-6, SEP-24 and SEP-8 for the same asset. If that's the case, we should still show the dropdown with SEP-6 and SEP-24 for that asset even if that's a regulated asset (SEP-8), right?

I think the only change we need is to add the `!isUntrusted &&` condition before rendering the `SEP-8 send` option, i.e. update:
```ts
            {asset.supportedActions?.sep8 && (
              <option value={AssetActionId.SEP8_SEND_PAYMENT}>
                SEP-8 Send
              </option>
            )}
```
to
```ts
            {!isUntrusted && asset.supportedActions?.sep8 && (
              <option value={AssetActionId.SEP8_SEND_PAYMENT}>
                SEP-8 Send
              </option>
            )}
```

WDYT?

quietbits: 👍 Sure, we can add that for clarity. Line 60 check should prevent it from being an untrusted asset for SEP-8 at that point, I think.

marcelosalloum (approved):
LGTM!

I created an issue to normalize how we display the dropdown options before adding the trustline: https://github.com/stellar/stellar-demo-wallet/issues/195

Issue #165, opened 2021-05-21
Feature Request: Asset suggestions in the "Add Asset" modeal


tomerweller:
Some knows assets on Stellar testnet could help users get started. Specifically SRT and USDC come to mind.

shanzzam:
@lydiat Easy enough to add a short "test with..." blurb when we're on testnet, breaking it down by related SEP. Larger convo is if it makes sense to have more visible How To on the home page.

PR #206, opened 2021-07-16
Show a list of suggested assets

marcelosalloum:
### What

Provide known assets from the Stellar testnet to help users get started:
* `USDC:GBBD47IF6LWK7P7MDEVSCWR7DPUWV3NY3DTQEVFL4NAT4AQH3ZLLFLA5`
* `SRT:testanchor.stellar.org`
* `MYASSET:sep8-server.dev.stellar.org`

Close stellar/stellar-demo-wallet#165

### Demo

https://user-images.githubusercontent.com/1952597/125978240-a9a0518a-bcae-4e88-9d62-a3dca00a29e3.mov

### DIFF
--- a/src/components/AddPresetAsset.tsx
+++ b/src/components/AddPresetAsset.tsx
@@ -61,7 +61,6 @@ export const AddPresetAsset = ({ onClose }: { onClose: () => void }) => {
 
           if (!(homeDomain || issuerPublicKey)) {
             const errorMsg = `Home domain OR issuer public key is required with asset code ${assetCode}`;
-            log.error({ title: errorMsg });
             throw new Error(errorMsg);
           }
 
@@ -98,14 +97,17 @@ export const AddPresetAsset = ({ onClose }: { onClose: () => void }) => {
 
   const renderAssetRow = (asset: presetAsset) => {
     const { homeDomain, issuerPublicKey: assetIssuer } = asset;
+
     const assetId = getAssetId(asset);
+    const networkUrl = getNetworkConfig(settings.pubnet).url.replace(
+      "https:",
+      "",
+    );
+
+    // if no home domain is provided, use horizon's /account endpoint:
     const issuerLink =
       (homeDomain && `//${homeDomain}/.well-known/stellar.toml`) ||
-      (assetIssuer &&
-        `${getNetworkConfig(settings.pubnet).url.replace(
-          "https:",
-          "",
-        )}/accounts/${assetIssuer}`);
+      (assetIssuer && `${networkUrl}/accounts/${assetIssuer}`);
     const displayLink =
       homeDomain || (assetIssuer && shortenStellarKey(assetIssuer));
 
--- a/src/components/AddPresetAsset.tsx
+++ b/src/components/AddPresetAsset.tsx
@@ -41,37 +41,17 @@ export const AddPresetAsset = ({ onClose }: { onClose: () => void }) => {
   }, [allAssets]);
 
   useEffect(() => {
-    setTimeout(() => {
-      if (!isValidating && untrustedAssets.status === ActionStatus.SUCCESS) {
-        onClose();
-      }
-    }, 100);
+    if (untrustedAssets.status === ActionStatus.SUCCESS) {
+      onClose();
+    }
 
     if (untrustedAssets.errorString) {
       setErrorMessage(untrustedAssets.errorString);
     }
-  }, [
-    isValidating,
-    onClose,
-    untrustedAssets.errorString,
-    untrustedAssets.status,
-  ]);
+  }, [onClose, untrustedAssets.errorString, untrustedAssets.status]);
 
   const getAssetId = (asset: presetAsset) =>
     `${asset.assetCode}:${asset.homeDomain || asset.issuerPublicKey}`;
-
-  const getAssetsToAdd = () => {
-    const assetsToAdd: presetAsset[] = [];
-
-    presetAssets.forEach((pAsset) => {
-      const assetId = getAssetId(pAsset);
-      if (checkedAssets[assetId]) {
-        assetsToAdd.push(pAsset);
-      }
-    });
-
-    return assetsToAdd;
-  };
 
   const handleOnChangeCheckbox = (asset: presetAsset) => {
     const updatedCheckedAssets = JSON.parse(JSON.stringify(checkedAssets));
@@ -123,6 +103,9 @@ export const AddPresetAsset = ({ onClose }: { onClose: () => void }) => {
     setIsValidating(false);
   };
 
+  const isPending =
+    isValidating || untrustedAssets.status === ActionStatus.PENDING;
+
   const renderAssetRow = (asset: presetAsset) => {
     const { homeDomain, issuerPublicKey: assetIssuer } = asset;
     const assetId = getAssetId(asset);
@@ -145,7 +128,7 @@ export const AddPresetAsset = ({ onClose }: { onClose: () => void }) => {
           onChange={() => {
             handleOnChangeCheckbox(asset);
           }}
-          disabled={isValidating}
+          disabled={isPending}
         />
         <div>
           <div className="PresetAssetCode">{asset.assetCode}</div>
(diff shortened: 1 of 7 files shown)


src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,180 @@
...
+    if (untrustedAssets.errorString) {
+      setErrorMessage(untrustedAssets.errorString);
+    }
+  }, [onClose, untrustedAssets.errorString, untrustedAssets.status]);
+
+  const getAssetId = (asset: presetAsset) =>
+    `${asset.assetCode}:${asset.homeDomain || asset.issuerPublicKey}`;
+
+  const handleOnChangeCheckbox = (asset: presetAsset) => {
+    const updatedCheckedAssets = JSON.parse(JSON.stringify(checkedAssets));
```
piyalbasu, line 57: I may be wrong, but since JSON.parse and JSON.stringify are inverses of each other, wouldn't `JSON.parse(JSON.stringify(checkedAssets))` be the same as just `checkedAssets`?

quietbits, line 57: `checkedAssets` is a local state variable we create, so we might not need these extra checks here? I think it might be safe to assume it is in the format we want it. Or make sure it is when we create it.

To make a clean copy of the list, we could do `[...checkedAssets]`.

src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+    try {
+      const validatedAssetValues = await Promise.all(
+        assetList.map(async (pAsset) => {
+          const { assetCode, issuerPublicKey, homeDomain } = pAsset;
+
+          if (!(homeDomain || issuerPublicKey)) {
+            const errorMsg = `Home domain OR issuer public key is required with asset code ${assetCode}`;
+            log.error({ title: errorMsg });
+            throw new Error(errorMsg);
```
quietbits, line 75: We might not need `log.error({ title: errorMsg });` here if we're logging it in `catch`.

src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,180 @@
...
+          }
+
+          const asset = await getValidatedUntrustedAsset({
+            assetCode,
+            homeDomain,
+            issuerPublicKey,
+            accountBalances: account.data?.balances,
+            networkUrl: getNetworkConfig(settings.pubnet).url,
```
quietbits, line 83: Do we need to use pubnet here?

src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+      const newSearchQ = searchParam.update(
+        SearchParams.UNTRUSTED_ASSETS,
+        validatedAssetValues,
+      );
+      history.push(newSearchQ);
+    } catch (e) {
+      const errorMsg = getErrorMessage(e);
+      log.error({ title: errorMsg });
+      setErrorMessage(errorMsg);
```
quietbits, line 98: If one of the assets fail, we quit the whole thing, correct?

piyalbasu (approved):

src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,180 @@
...
+  const renderAssetRow = (asset: presetAsset) => {
+    const { homeDomain, issuerPublicKey: assetIssuer } = asset;
+    const assetId = getAssetId(asset);
+    const issuerLink =
+      (homeDomain && `//${homeDomain}/.well-known/stellar.toml`) ||
+      (assetIssuer &&
+        `${getNetworkConfig(settings.pubnet).url.replace(
+          "https:",
+          "",
+        )}/accounts/${assetIssuer}`);
```
quietbits, line 118: Could we add a comment here for clarity? 😊 That we are using accounts endpoint data if there is no home domain, correct?

Also, same as earlier about pubnet.

src/constants/presetAssets.ts
```diff
@@ -0,0 +1,20 @@
+export interface presetAsset {
+  assetCode: string;
+  homeDomain?: string;
+  issuerPublicKey?: string;
+}
```
quietbits, line 5: We could move this to `types` file.

src/helpers/searchParam.ts
```diff
@@ -7,7 +7,7 @@ import { SearchParams, SearchParamAsset, StringObject } from "types/types.d";
 
 const update = (
   param: SearchParams,
-  value: string,
+  value: string | string[],
```
quietbits, line 10: I don't remember exactly what, but I think I ran into some issues when using arrays in URL params. My idea was that we'd pass comma-separated strings here. That didn't work for this case?

src/components/AddPresetAsset.tsx
marcelosalloum: Yup! This `pubnet` is a boolean, it's set to true for pubnet and false for testnet:
https://github.com/stellar/stellar-demo-wallet/blob/790e8ba3812ee9bb5ebf73ffbce19f8ea2e674b8/src/types/types.d.ts#L130

shanzzam: Can we want to only allow this for testnet? I don't think it makes sense to run this on mainnet.

shanzzam:
UX looks good. I added one comment above regarding use on testnet only.

src/helpers/searchParam.ts
marcelosalloum: Good idea, I hadn't thought about that! I'll try to revert these changes and make the PR shorter...

marcelosalloum: Done in https://github.com/stellar/stellar-demo-wallet/pull/206/commits/c89e33faf5c9a4e84945ca0ddaf02d05b7e6c838.

src/components/AddPresetAsset.tsx
marcelosalloum: Done in https://github.com/stellar/stellar-demo-wallet/pull/206/commits/3a63377862a8265c9bc0e413fcd018ed7a13d9f0.

src/components/AddPresetAsset.tsx
marcelosalloum: Good point! Fixed in https://github.com/stellar/stellar-demo-wallet/pull/206/commits/3a63377862a8265c9bc0e413fcd018ed7a13d9f0.

src/components/AddPresetAsset.tsx
quietbits: My bad, @marcelosalloum , I read it wrong after being away from this project for a while. Maybe we should use a boolean type of name for this (`isPubnet`). 🤔 Not related to this PR, of course.

src/components/AddPresetAsset.tsx
marcelosalloum: TIL! I still had this old approach in mind for creating a copy, thanks for pointing out the leaner alternative!

Also, I stopped using `useState` for `checkedAssets` in https://github.com/stellar/stellar-demo-wallet/pull/206/commits/cf431f7cac358ad36618174de339e4ea1eb2f5db, thanks to your suggestion 🙏 .

src/constants/presetAssets.ts
marcelosalloum: Done in https://github.com/stellar/stellar-demo-wallet/pull/206/commits/839cc6b2a4d0a8dcb694415ab69c459b5df69f47.

src/components/AddPresetAsset.tsx
marcelosalloum: Yes, that's correct. The risk is low, considering those assets are hardcoded but could happen right after a testnet reset.

src/components/AddPresetAsset.tsx
marcelosalloum: Good point, Shannon! With [this commit](https://github.com/stellar/stellar-demo-wallet/pull/206/commits/65ee3f6ee9589848ab0773d4349b01dbf468d325), we now hide that button if in pubnet.

src/components/AddPresetAsset.tsx
```diff
@@ -0,0 +1,171 @@
...
+            networkUrl: getNetworkConfig(settings.pubnet).url,
+          });
+
+          return `${asset.assetCode}:${asset.assetIssuer}`;
+        }),
+      );
+
+      const newSearchQ = searchParam.update(
+        SearchParams.UNTRUSTED_ASSETS,
+        validatedAssetValues.join(","),
```
quietbits, line 81: 👏 I like this better.

quietbits (approved):

PR #215, opened 2021-07-26
Dark mode

quietbits:
image


src/App.scss
```diff
@@ -1,43 +1,49 @@
...
-}
+#root {
+  --layout-window-width-min: 800px;
+  --layout-window-width-max: 1296px;
 
-body {
-  overflow-y: hidden;
-}
+  // TODO: fix this in SDS
+  overflow-x: auto !important;
```
marcelosalloum, line 6: Do you have a ticket for this TODO? Are you tracking it somehow?

src/components/CopyWithText.tsx
```diff
@@ -28,7 +28,7 @@ export const CopyWithText = ({
     <>
       {children}
       <CopyToClipboard text={textToCopy} onCopy={handleCopy}>
-        <TextButton>{inProgress ? "Copied" : "Copy"}</TextButton>
+        <TextLink role="button">{inProgress ? "Copied" : "Copy"}</TextLink>
```
marcelosalloum, line 31: Instead of using a string for the role in `role="button"` we could use an enum. This is probably something for SDS though.

marcelosalloum (commented):
Excited to see the dark theme landing!

I've found an issue with the light theme colors in:
Screen Shot 2021-07-26 at 17 38 50

src/App.scss
quietbits: Yes, I have a list where I'm tracking all this. 👍

src/components/CopyWithText.tsx
quietbits: 🤔 Hmm, I'm not sure we need to create an enum for roles, because the values are coming from [HTML](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques).

src/components/Header.tsx
```diff
@@ -14,19 +14,22 @@ export const Header = () => {
...
-          <TextButton onClick={() => setModalVisible(true)}>
-            Sign out
-          </TextButton>
-        )}
-      </div>
+    <>
+      <Layout.Header
+        projectTitle="Demo Wallet"
+        onSignOut={
+          account.isAuthenticated ? () => setModalVisible(true) : undefined
```
piyalbasu, line 21: pretty NIT: but I think it's best practice to use `null` here instead of `undefined` because `null` denotes that you are explicitly saying "do nothing"

piyalbasu (approved):

src/components/Header.tsx
quietbits: The idea here is to make `onSignOut` an optional prop, so I thought that `undefined` would be clear, but looks like it's not. 😅 Should we use conditional prop spread instead? I wouldn't want to add an extra type `null` to the types [here](https://github.com/stellar/stellar-design-system/blob/main/%40stellar/design-system/src/components/Layout/index.tsx#L39), as it should be just an optional prop with a function type.

quietbits:
> Excited to see the dark theme landing!
> 
> I've found an issue with the light theme colors in:
> Screen Shot 2021-07-26 at 17 38 50

@marcelosalloum The dark mode is not completed yet, there are plenty of issues at the moment. Very soon. 🤓

src/components/Header.tsx
marcelosalloum: Would this work better?
```ts
onSignOut={() => account.isAuthenticated && setModalVisible(true)}
```

piyalbasu: Makes sense. Yeah, I think a conditional prop spread is the way to go 👍

quietbits: 👍 Updated in [this PR](https://github.com/stellar/stellar-demo-wallet/pull/216) (I don't want to make changes to this one since it will be the base).

PR #216, opened 2021-07-27
Dark mode: modals + CopyText + Logs + Json

quietbits:
- All modals should be using the `Modal` component from SDS (in some cases content might still need to be fixed, but this is coming in another PR).
- `CopyText` and `Tooltip` components added (will be added to SDS a bit later) and removed previous local ones.
- Removed local SVG icons, use SDS icons instead.
- Logs should be fully updated now.
- Updated `Json` component (using this everywhere instead of `ReactJson` with local updates).
- Fixed main panel scrolling and conditional prop in `Header`.

image

image


src/components/Tooltip/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+  className?: string;
+}
+
+export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
+  (props, ref) => {
+    const { isVisible, children, className } = props;
+
+    return (
+      <div
+        className={`Tooltip ${className ?? ""}`}
```
piyalbasu, line 17: very NIT: using `??` implies you want to use `className` if it is equal to `0` or `""`, but I think you probably want `className || ""` here

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

piyalbasu (approved):

src/components/Tooltip/index.tsx
quietbits: I think it's OK to allow a `0` string in this case. Thanks for pointing it out! 👍

PR #220, opened 2021-07-28
Details tooltip

quietbits:
- `DetailsTooltip` component
- Use SDS components instead of local
- Ignoring some TypeScript errors until SDS is updated (will do it after all changes are good locally)


src/components/AccountInfo.tsx
```diff
@@ -91,34 +89,42 @@ export const AccountInfo = () => {
...
-                  tooltipText="If you performed account actions elsewhere, like in the
+                <DetailsTooltip
+                  details="If you performed account actions elsewhere, like in the
                   Stellar Laboratory, click here to update."
                 >
-                  Refresh account
-                </TextButton>
+                  <TextLink
+                    onClick={handleRefreshAccount}
+                    disabled={account.status === ActionStatus.PENDING}
```
piyalbasu, line 123: might be good to cache `account.status === ActionStatus.PENDING` as `isPending` since it's being checked twice

src/components/AddAsset.tsx
```diff
@@ -112,60 +113,88 @@ export const AddAsset = ({ onClose }: { onClose: () => void }) => {
...
-          label="Asset code"
+          // TODO: change type in SDS
+          // @ts-ignore
+          label={
+            <DetailsTooltip
+              details={
+                <>
+                  Assets are identified by 1) their code and 2) either a home
+                  domain or the public key of the issuing account.{" "}
+                  <TextLink href="https://developers.stellar.org/docs/issuing-assets/publishing-asset-info/">
```
piyalbasu, line 124: Out of scope for this PR - but at some point it might be good to create a const called `DEVELOPER_DOCS_URL` or something so we don't need to write out `https://developers.stellar.org/docs/` for every url

piyalbasu (approved):

PR #223, opened 2021-07-29
SEP-6 deposit: send amount param in request

quietbits:
image


JakeUrban (approved):
Looks like it's working!

Issue #219, opened 2021-07-28
Error persists in modal dialoge 


shanzzam:
### What did you do?
I tested the no_op_error issue to replicate the error state by sending SRT to account that didn't have a trustline to SRT.
After getting the error, I then went to send XML by a normal on-chain payment. 
After that, I again went to do another on-chain send with SRT.

### What did you expect to see?
I expected to see a Send dialog with no errors. 

### What did you see instead?
The no_op_error persisted in the Send modal. See video.

https://user-images.githubusercontent.com/5062622/127242804-f70c9f35-6910-4184-a8f0-0c6c3dc36361.mov

Issue #217, opened 2021-07-27
SEP-6: Amount is not sent in deposit request


JakeUrban:
When I go through a SEP-6 flow and enter an amount to deposit, no `amount` field is included in the `GET /deposit` request to the anchor. The test anchor defaults to a value of 103 if no `amount` value was provided in the request, but this will vary per-anchor.

We should update the wallet to send the `amount` field if an amount was specified in the form.

Issue #214, opened 2021-07-26
Inconsistent home domain parameter 


meiyib:
### What version are you using?
Version 1.2.0

### What did you do?
When checking the configuration of the demo wallet, under Query String Parameters, the requested 'homeDomain' parameter format is different from SEP-10 request parameters 'home_domain'.

### What did you expect to see?
I expect to see 'home_domain' parameter, so it's consistent with SEP-10 guidelines. 

### What did you see instead?
It's currently written as 'homeDomain'

<img width="530" alt="Screen Shot 2021-07-26 at 9 18 37 AM" src="https://user-images.githubusercontent.com/53131684/127038628-a8336935-6f35-4046-96db-dfe210d8bb87.png">
<img width="539" alt="Screen Shot 2021-07-26 at 9 18 50 AM" src="https://user-images.githubusercontent.com/53131684/127038633-50612ac8-8cf1-4004-af5d-cef376db3f70.png">

PR #226, opened 2021-07-29
Release/1.2.1

quietbits:
- Dark Mode Support comes to the Demo Wallet
- Added more error codes
- Updated Sentry errors and analytics with a standard library. This will help us target and fix issues quicker, as well as provide feedback on the features that are actively used, allowing us to prioritize features.

- Use the correct home_domain syntax for SEP-10 compliance: Issue #214 
- SEP-06 deposit flow updated to send the correct amount in the request: Issue #217
- Fixed an issue where the error message would remain in the dialog box under certain conditions: Issue #219


piyalbasu (approved):

marcelosalloum (approved):
Can you add the description from the previous PR here too?

PR #240, opened 2021-09-10
Update docker image build timestamp format

jacekn:
Golang time parser is not compatible with format rendered by the GNU date,
even though both claim to be RFC3339. This change replaces space separator
between date and time with "T" which will allow golang to parse it.
It also switches from `--utc` to `-u` that's supported by both GNU and BSD
date commands.


tprobinson (approved):
LGTM

PR #252, opened 2021-11-03
Update README.md for deposits with the reference server

JakeUrban:
An additional step was added to the reference server's interactive flow for selecting the off-chain asset to provide to the anchor (which is just for demonstration since the reference server doesn't actually require off-chain value to be provided). If "USD" is selected on this form, the user is asked to confirm the fictional exchange rate between USD and SRT.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ right of the screen will show every network call.
 7. Click “Skip confirmation” - skipping it won't be possible in live
    integrations but helps the process move ahead in the demo.
 8. Select the asset you would like to provide to the anchor. Note that there
-   is no real asset you actually need to provide off-chain, its just for 
+   is no real asset you actually need to provide off-chain, it's just for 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -65,9 +65,15 @@ right of the screen will show every network call.
    real, but the interface will want a valid email.
 7. Click “Skip confirmation” - skipping it won't be possible in live
    integrations but helps the process move ahead in the demo.
-8. Enter a number into the amount and click "Submit".
-9. Leave the pop-up window open while you wait to see the deposit of SRT go
-   through - you can close when you see “Status” is complete and you have SRT.
+8. Select the asset you would like to provide to the anchor. Note that there
+   is no real asset you actually need to provide off-chain, its just for
```
lydiat, line 69: it's :)

lydiat (approved):

PR #261, opened 2021-12-22
SEP-10: get home domain from TOML server URL

quietbits:
For SEP-10 home domain, we were incorrectly passing the URL where the TOML file is hosted, instead of getting a transaction-specific server URL from the TOML file itself. This PR should fix that. :crossed_fingers: 

Update SEP-6. SEP-24, and SEP-31 transactions.


quietbits:
The current behavior is correct, closing the PR.

PR #275, opened 2022-04-15
SEP-31: make sure we get the amount_in from GET /transactions before …

marcelosalloum:
### What

Make sure we get the `amount_in` from `GET /transactions` before sending the stellar payment.

### DIFF
--- a/packages/demo-wallet-client/src/ducks/sep31Send.ts
+++ b/packages/demo-wallet-client/src/ducks/sep31Send.ts
@@ -324,15 +324,20 @@ export const submitSep31SendTransactionAction = createAsyncThunk<
       });
 
       // Poll transaction until ready
-      await pollTransactionUntilReady({
+      const getSep31Tx = await pollTransactionUntilReady({
         sendServer,
         transactionId: postResponse.transactionId,
         token,
       });
 
+      const amountIn = getSep31Tx?.transaction?.amount_in;
+      if (amountIn === undefined) {
+        throw new Error(`"amount_in" is missing from the GET /transaction response`);
+      }
+
       // Send payment
       await sendPayment({
-        amount: amount.amount,
+        amount: getSep31Tx!.transaction!.amount_in as string,
         assetCode,
         assetIssuer,
         receiverAddress: postResponse.receiverAddress,
--- a/packages/demo-wallet-shared/methods/sep31Send/pollTransactionUntilReady.ts
+++ b/packages/demo-wallet-shared/methods/sep31Send/pollTransactionUntilReady.ts
@@ -1,5 +1,5 @@
 import { log } from "../../helpers/log";
-import { TransactionStatus } from "../../types/types";
+import { TransactionStatus, Sep31GetTransaction } from "../../types/types";
 
 export const pollTransactionUntilReady = async ({
   sendServer,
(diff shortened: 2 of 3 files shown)


packages/demo-wallet-client/src/ducks/sep31Send.ts
```diff
@@ -324,15 +324,20 @@ export const submitSep31SendTransactionAction = createAsyncThunk<
       });
 
       // Poll transaction until ready
-      await pollTransactionUntilReady({
+      const getSep31Tx = await pollTransactionUntilReady({
```
quietbits, line 327: TypeScript is not happy in my VS Code. Can you please try to set a type on `getSep31Tx`? Also, current name sounds like it is a function, maybe we can name it `sep31Tx`?

image

packages/demo-wallet-client/src/ducks/sep31Send.ts
```diff
@@ -324,15 +324,20 @@ export const submitSep31SendTransactionAction = createAsyncThunk<
...
 
+      const amountIn = getSep31Tx?.transaction?.amount_in;
+      if (amountIn === undefined) {
+        throw new Error(`"amount_in" is missing from the GET /transaction response`);
+      }
+
       // Send payment
       await sendPayment({
-        amount: amount.amount,
+        amount: getSep31Tx!.transaction!.amount_in as string,
```
quietbits, line 340: We could use `amountIn` here.

packages/demo-wallet-shared/methods/sep31Send/pollTransactionUntilReady.ts
```diff
@@ -32,14 +33,16 @@ export const pollTransactionUntilReady = async ({
...
 
     // eslint-disable-next-line no-await-in-loop
-    const resultJson = await result.json();
+    resultJson = await result.json();
     log.response({
       title: `GET \`/transactions/${transactionId}\``,
       body: resultJson,
     });
-    transactionStatus = resultJson.transaction.status;
+    transactionStatus = resultJson.transaction?.status;
```
quietbits, line 41: Do we want to set a default status here or no status is acceptable?

marcelosalloum:
Thank you for your review.

I'm converting this to draft for now while the SEP change https://github.com/stellar/stellar-protocol/pull/1199 is not complete.

packages/demo-wallet-shared/methods/sep31Send/pollTransactionUntilReady.ts
marcelosalloum: There's no need fr a default value because all we want in this part of the code is to verify if the status is `TransactionStatus.PENDING_SENDER`. Here's the full code snippet for reference:
https://github.com/stellar/stellar-demo-wallet/blob/9c7e3f3f99a6563df26b95a385c883a300356769/packages/demo-wallet-shared/methods/sep31Send/pollTransactionUntilReady.ts#L19-L44

packages/demo-wallet-client/src/ducks/sep31Send.ts
marcelosalloum: https://github.com/stellar/stellar-demo-wallet/pull/275/commits/5bf0e1caac9386e27bf1fa028723fd0aa14e52f1

packages/demo-wallet-client/src/ducks/sep31Send.ts
marcelosalloum: https://github.com/stellar/stellar-demo-wallet/pull/275/commits/5bf0e1caac9386e27bf1fa028723fd0aa14e52f1

quietbits (approved):

PR #281, opened 2022-05-25
Custodial mode for SEP-24 deposit and withdrawal

quietbits:
- [x] Add custodial inputs to UI only for SEP-24 deposit and withdrawal
- [x] Add `custodial` to Redux store + handle data update
- [x] SEP-24 deposit: update SEP-10 to support custodial fields
- [x] SEP-24 deposit: add generated `memo`
- [x] SEP-24 withdrawal: update SEP-10 to support custodial fields
- [x] ~SEP-24 withdrawal: add generated `memo`~

image


packages/demo-wallet-client/src/ducks/sep24DepositAsset.ts
```diff
@@ -51,6 +59,16 @@ export const depositAssetAction = createAsyncThunk<
...
 
+    // This is unlikely
+    if (
+      custodialIsEnabled &&
+      !(custodialSecretKey && custodialPublicKey && custodialMemoId)
+    ) {
+      throw new Error(
+        "Custodial mode requires secret key, public key, and memo ID",
+      );
+    }
```
JakeUrban, line 70: Is it possible? Doesn't the custodial component ensure these values are collected before initiating a deposit?

packages/demo-wallet-shared/methods/sep24/interactiveDepositFlow.ts
```diff
@@ -15,15 +28,19 @@ export const interactiveDepositFlow = async ({
...
   const formData = new FormData();
-  const postDepositParams = {
+  const postDepositParams: DepositParams = {
     asset_code: assetCode,
     account: publicKey,
     lang: "en",
     claimable_balance_supported: claimableBalanceSupported.toString(),
+    ...(isCustodialMode
+      ? { memo: Math.floor(Math.random() * 100).toString(), memo_type: "id" }
+      : {}),
```
JakeUrban, line 43: Ideally, we throw an error in the logs if the anchor does not use the memo generated here in the transaction that completes the deposit. Is that an easy lift?

JakeUrban: Or maybe not an error, but a warning?

JakeUrban (approved):
This looks great, thanks for getting this out so quickly. I have one question and one request.

packages/demo-wallet-client/src/ducks/sep24DepositAsset.ts
quietbits: This should never happen, adding here just in case something goes wrong when updating the state.

packages/demo-wallet-shared/methods/sep24/interactiveDepositFlow.ts
quietbits: Added the check when we start polling transaction status. Here's an example of the warning.

image

JakeUrban (approved):

PR #286, opened 2022-06-07
SEP-24: add optional SEP-9 fields to deposit and withdrawal

quietbits:
- Created `extra` param in store for various things (like SEP-9 fields)
- No validation or formatting added to the SEP-9 fields (submitted as entered)
- Updated SEP-24 deposit and withdrawal flow to accept SEP-9 fields if provided

image


quietbits:
Added memo

image

JakeUrban:
Hey Iveta, can you describe the application logic here? How does the demo wallet know if the anchor accepts SEP-9 fields in the deposit and withdraw requests, and how does it know which fields it can provide? Does the user of the demo wallet interactively select the fields they want to provide?

quietbits:
Hey, @JakeUrban ! As we discussed offline, we'll allow users to select SEP-9 fields to submit dynamically, as you suggested above.

JakeUrban (approved):
Looks great!

Ifropc (approved):
Looks great, thanks @quietbits !

PR #307, opened 2023-03-27
[PARTNER-28] Don't redirect to more_info in demo_wallet too early

Ifropc:
Demo wallet has an invalid assumption that the transaction status won’t change until the user is done with the KYC / transaction info collection.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ yarn-error.log*
 
 .eslintcache
 .nvmrc
+.idea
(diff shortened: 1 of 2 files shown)


JakeUrban (commented):
I tested this out with testanchor.stellar.org and it didn't open a second popup, maybe because the main window was already on the more info url?

I'm thinking, maybe the demo wallet shouldn't do anything on status change. I think we can just poll the `GET /transactions/:id` endpoint and put updates in the demo wallet logs, and let the user decide when to refresh.

Ifropc:
> @Ifropc How should I test this? I tried SEP-24 and SEP-6 deposits with SRT asset (testanchor.stellar.org). Both were completed, and the logs look fine.

It's for SEP-24 deposit. I asked one of our partners (who were affected by redirect) to test it, they were happy with no redirect

quietbits (approved):

PR #310, opened 2023-04-27
Add message to the status page

Ifropc:


packages/demo-wallet-shared/methods/sep24/pollDepositUntilComplete.ts
```diff
@@ -54,7 +54,7 @@ export const pollDepositUntilComplete = async ({
         });
       }
       log.instruction({
-        title: `Transaction \`${transactionId}\` is in \`${transactionJson.transaction.status}\` status`,
+        title: `Transaction \`${transactionId}\` is in \`${transactionJson.transaction.status}\` status. Message: \`${transactionJson.transaction.message}\``,
```
JakeUrban, line 57: While we're at it, can we log the transaction body whenever the status changes?

JakeUrban (approved):

packages/demo-wallet-shared/methods/sep24/pollDepositUntilComplete.ts
JakeUrban: Lets do this for withdraws and sends too.

packages/demo-wallet-shared/methods/sep24/pollDepositUntilComplete.ts
```diff
@@ -54,7 +54,8 @@ export const pollDepositUntilComplete = async ({
         });
       }
       log.instruction({
-        title: `Transaction \`${transactionId}\` is in \`${transactionJson.transaction.status}\` status`,
+        title: `Transaction \`${transactionId}\` is in \`${transactionJson.transaction.status}\` status. \n
+        Transaction body: \`\`\`${JSON.stringify(transactionJson.transaction)}\`\`\``,
```
JakeUrban, line 58: I think we want to use `log.response` for the transaction body because it has better styling.

PR #320, opened 2023-08-15
Replace CRA with Webpack + Node 18

quietbits:


packages/demo-wallet-client/webpack.common.js
```diff
@@ -0,0 +1,160 @@
...
+        },
+        "vendor-stellar": {
+          test: /[\\/]node_modules[\\/](@stellar|stellar)[\\/]/,
+          name: "vendor-stellar",
+          chunks: "all",
+        },
+        "vendor-wallets": {
+          test: /[\\/]node_modules[\\/](@albedo|@ledgerhq|trezor)[\\/]/,
+          name: "vendor-wallets",
+          chunks: "all",
```
piyalbasu, line 36: I like this organization of node_modules! From the docs, my assumption this is a way to make sure we don't duplicate any of these libs (thus reducing bloat). Is that what your aim is here? Curious if there are any other benefits I'm missing

piyalbasu (approved):

packages/demo-wallet-client/webpack.common.js
quietbits: If I remember correctly this is to make the generated file smaller by adding these only once. It was initially done in the Account Viewer in [this PR](https://github.com/stellar/account-viewer-v2/pull/387). Looks like it was a lot of trial and error to get it right. :sweat_smile:

PR #327, opened 2023-10-20
SEP-12 KYC: use field type for input

quietbits:
image

### DIFF
--- a/packages/demo-wallet-client/src/components/KycFieldInput.tsx
+++ b/packages/demo-wallet-client/src/components/KycFieldInput.tsx
@@ -23,7 +23,6 @@ const getInputType = (type: KycFieldType) => {
     case "number":
       return "number";
     case "string":
-      return;
     default:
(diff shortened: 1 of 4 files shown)


JakeUrban (commented):
If I remember correctly we were submitting KYC via JSON, but if we have files we'll need to submit using form-data. Can you verify we have that support?

quietbits:
@JakeUrban We are [submitting KYC fields](https://github.com/stellar/stellar-demo-wallet/blob/master/packages/demo-wallet-shared/methods/sep12/putSep12FieldsRequest.ts#L30:L41) as form data. It worked when I tested SEP-6 deposit/withdrawal and SEP-31, but I'm not sure if SRT validates the image input.

JakeUrban:
It validates and deserializes the content in the request so I think we're good then.

JakeUrban (approved):

PR #330, opened 2023-11-09
SEP-6 updates

quietbits:
Deposit
[sep6-demo-deposit.webm](https://github.com/stellar/stellar-demo-wallet/assets/7346473/48d70210-f0df-4abe-82f5-5c0fd8dc22fc)

Withdrawal
[sep6-demo-withdrawal.webm](https://github.com/stellar/stellar-demo-wallet/assets/7346473/b908779a-ba15-45b3-ac18-14fb97be372d)

### DIFF
--- a/packages/demo-wallet-client/src/components/KycFieldInput.tsx
+++ b/packages/demo-wallet-client/src/components/KycFieldInput.tsx
@@ -38,6 +38,7 @@ export const KycFieldInput = ({
   onChange,
   isRequired,
 }: KycFieldInputProps) => {
+  // isRequired param allows to override optional setting returned from the API
   const _isRequired = isRequired || !input.optional;
   const label = `${input.description}${_isRequired ? "" : " (optional)"}`;
 
--- a/packages/demo-wallet-client/src/components/KycFieldInput.tsx
+++ b/packages/demo-wallet-client/src/components/KycFieldInput.tsx
@@ -1,4 +1,4 @@
-import { Input } from "@stellar/design-system";
+import { Input, Select } from "@stellar/design-system";
 
 type KycFieldType = "string" | "binary" | "number" | "date";
 
@@ -6,12 +6,16 @@ export type KycField = {
   type: KycFieldType;
   description: string;
   optional?: boolean;
+  choices?: string[];
 };
 
 type KycFieldInputProps = {
   id: string;
   input: KycField;
-  onChange: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  onChange: (
+    event: React.ChangeEvent<HTMLInputElement | HTMLSelectElement>,
+  ) => void;
+  isRequired?: boolean;
 };
 
 const getInputType = (type: KycFieldType) => {
@@ -28,8 +32,31 @@ const getInputType = (type: KycFieldType) => {
   }
 };
 
-export const KycFieldInput = ({ id, input, onChange }: KycFieldInputProps) => {
-  const label = `${input.description}${input.optional ? " (optional)" : ""}`;
+export const KycFieldInput = ({
+  id,
+  input,
+  onChange,
+  isRequired,
+}: KycFieldInputProps) => {
+  const _isRequired = isRequired || !input.optional;
+  const label = `${input.description}${_isRequired ? "" : " (optional)"}`;
+
+  if (input.choices?.length) {
+    return (
+      <Select
+        key={id}
+        id={id}
+        label={label}
+        required={_isRequired}
+        onChange={onChange}
+      >
+        <option></option>
+        {input.choices.map((val) => (
+          <option value={val}>{val}</option>
+        ))}
+      </Select>
+    );
+  }
 
   return (
     <Input
--- a/packages/demo-wallet-client/src/components/Sep6/Sep6Deposit.tsx
+++ b/packages/demo-wallet-client/src/components/Sep6/Sep6Deposit.tsx
@@ -21,7 +21,7 @@ import {
 } from "ducks/sep6DepositAsset";
 import { useRedux } from "hooks/useRedux";
 import { AppDispatch } from "config/store";
-import { ActionStatus, SepInstructions } from "types/types";
+import { ActionStatus } from "types/types";
 
 export const Sep6Deposit = () => {
   const { sep6DepositAsset } = useRedux("sep6DepositAsset");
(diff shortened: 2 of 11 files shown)


packages/demo-wallet-client/src/components/Sep6/Sep6Deposit.tsx
```diff
@@ -284,7 +323,7 @@ export const Sep6Deposit = () => {
   if (sep6DepositAsset.status === ActionStatus.CAN_PROCEED) {
     return (
       <Modal visible onClose={handleClose} parentId={CSS_MODAL_PARENT_ID}>
-        <Modal.Heading>SEP-6 Deposit Success</Modal.Heading>
+        <Modal.Heading>SEP-6 Deposit Details</Modal.Heading>
```
philipliu, line 326: Should we still display "Deposit Success" when the transaction completes?

philipliu (approved):
I only reviewed this at a high level but the changes LGTM!

JakeUrban:
What anchor are we testing against? https://testanchor.stellar.org doesn't have the protocol changes we've made

quietbits:
> What anchor are we testing against? https://testanchor.stellar.org doesn't have the protocol changes we've made

@JakeUrban we're testing it with Anchor Platform (running it locally).

JakeUrban:
I see, any chance we can get a demo video? I don't have an anchor platform setup for SEP-6 locally

quietbits:
Yes, I'll post videos on this PR once the PR is done.

packages/demo-wallet-client/src/components/Sep6/Sep6Withdraw.tsx
```diff
@@ -245,46 +286,58 @@ export const Sep6Withdraw = () => {
...
-          </div>
-
-          <Input
-            id="withdraw-amount"
-            label="Amount to Withdraw"
-            required
-            onChange={handleAmountFieldChange}
-          />
-          {withdrawResponse.min_amount || withdrawResponse.max_amount ? (
+          {withdrawResponse.account_id ? (
```
quietbits, line 300: `account_id` is optional in `withdraw` response now.

packages/demo-wallet-client/webpack.dev.js
```diff
@@ -8,7 +8,7 @@ module.exports = merge(common, {
   devServer: {
     static: path.join(__dirname, "build"),
     historyApiFallback: true,
-    port: 3000,
+    port: 4000,
```
quietbits, line 11: `3000` port was clashing with the Anchor Platform, so I changed it to something else to avoid that.

packages/demo-wallet-client/src/components/KycFieldInput.tsx
```diff
@@ -28,8 +32,31 @@ const getInputType = (type: KycFieldType) => {
...
 
-export const KycFieldInput = ({ id, input, onChange }: KycFieldInputProps) => {
-  const label = `${input.description}${input.optional ? " (optional)" : ""}`;
+export const KycFieldInput = ({
+  id,
+  input,
+  onChange,
+  isRequired,
+}: KycFieldInputProps) => {
+  const _isRequired = isRequired || !input.optional;
```
piyalbasu, line 41: what are the differences between `isRequired` and `input.optional`? Is there instances when they are out of sync? My understanding is the `!input.optional === isRequired`, but maybe that's not the case

quietbits: I should have clarified this. `isRequired` flag overrides `input.optional`, which we get from the API. We allow optional fields when they are initially required in KYC, but we make them all required when they are in `required_customer_info_updates` field requested later.

JakeUrban:
In the deposit video, it looks like we're collecting & sending KYC before we initiate the transaction via `GET /deposit`. How did we know which fields to pass? Realistically, the wallet doesn't know and would need to use a valid customer type defined by the anchor, which we're trying to avoid.

So, we should be initiating the transaction before collecting any KYC and then polling it to see what fields need to be collected. Does that match your understanding @philipliu ?

philipliu:
>In the deposit video, it looks like we're collecting & sending KYC before we initiate the transaction via GET /deposit. How did we know which fields to pass? Realistically, the wallet doesn't know and would need to use a valid customer type defined by the anchor to fetch the fields required, which we're trying to avoid.

The existing SEP-6 anchors require that the customer is in the ACCEPTED SEP-12 customer state with no type before the deposit is initiated. This PR does not touch this part of the flow but only handles KYC collection when the transaction moves into the `pending_customer_info_update` status and displays the deposit instructions. Collecting KYC upfront is not required by Anchor Platform and is no longer recommended going forward as of the recent stellar-anchor-tests change https://github.com/stellar/stellar-anchor-tests/pull/131.

> So, we should be initiating the transaction before collecting any KYC and then polling it to see what fields need to be collected. Does that match your understanding @philipliu ?

I think this makes sense. We can push the KYC collection to after the GET /deposit request. To not break existing anchors, we would need to handle the 403 NEEDS_INFO response and do the existing GET /customer, PUT /customer flow. This should initialize the transaction and we can continue to poll it to see if more fields are required. 

How does that sound, @JakeUrban?

Ifropc:
How do we get list of fields to submit for SEP-6 deposit, before the deposit started? Do we do it with using GET SEP-12 /customer?  Couldn't we just disable SEP-12 on the platform configuration, and demo wallet will just skip this step if SEP-12 not available?

JakeUrban:
> We can push the KYC collection to after the GET /deposit request. To not break existing anchors, we would need to handle the 403 NEEDS_INFO response and do the existing GET /customer, PUT /customer flow. This should initialize the transaction and we can continue to poll it to see if more fields are required.

Sounds perfect to me.

> Couldn't we just disable SEP-12 on the platform configuration, and demo wallet will just skip this step if SEP-12 not available?

If the anchor doesn't require KYC then yes, but realistically that won't happen. The wallet will need the SEP-12 endpoints for two reasons:

1. Fetch the descriptions & choices of the fields requested in the transaction record
2. Provide the values collected from the user to the anchor

Ifropc:
No I mean to not call SEP-12 before SEP-6 in the demo wallet. I think I was talking about the same thing Philip mentioned, now that I read it again

Ifropc (approved):

PR #335, opened 2024-01-04
Refactor asset data fetching and handling

quietbits:
Refactored to fetch asset data "on demand" instead of always fetching on-chain data, which resulted in unnecessary calls for assets with home domain override.

### Unfunded account

- [x]  Add untrusted asset
- [x]  Add untrusted asset with override
- [x]  Remove override from untrusted asset
- [x]  Create account with untrusted asset
- [x]  Create account with untrusted asset with override

### Funded account

- [x]  Add untrusted asset
- [x]  Add untrusted asset with override
- [x]  Remove override from untrusted asset
- [x]  Add override to untrusted asset
- [x]  Add trustline with untrusted asset
- [x]  Add trustline with untrusted asset with override
- [x]  Add override to trusted asset
- [x]  Remove override from trusted asset
- [x] SEP-6 deposit with trusted asset shows correct balance
- [x] SEP-6 deposit with untrusted asset shows correct balance
- [x] SEP-24 deposit with trusted asset and override shows correct balance
- [x] SEP-24 deposit with untrusted asset and override shows correct balance

### DIFF
--- a/packages/demo-wallet-client/src/ducks/allAssets.ts
+++ b/packages/demo-wallet-client/src/ducks/allAssets.ts
@@ -1,10 +1,16 @@
 import { createAsyncThunk, createSlice } from "@reduxjs/toolkit";
+import { Types } from "@stellar/wallet-sdk";
+
 import { RootState } from "config/store";
 import { accountSelector } from "ducks/account";
 import { assetOverridesSelector } from "ducks/assetOverrides";
 import { untrustedAssetsSelector } from "ducks/untrustedAssets";
+import { settingsSelector } from "ducks/settings";
 import { getErrorMessage } from "demo-wallet-shared/build/helpers/getErrorMessage";
 import { log } from "demo-wallet-shared/build/helpers/log";
+import { getAssetData } from "demo-wallet-shared/build/helpers/getAssetData";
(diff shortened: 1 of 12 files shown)


packages/demo-wallet-client/src/ducks/allAssets.ts
```diff
@@ -15,43 +21,62 @@ import {
...
   { rejectValue: RejectMessage; state: RootState }
->("allAssets/getAllAssetsAction", (_, { rejectWithValue, getState }) => {
-  const { assets } = accountSelector(getState());
+>("allAssets/getAllAssetsAction", async (_, { rejectWithValue, getState }) => {
+  const networkConfig = getNetworkConfig();
+
+  const { data } = accountSelector(getState());
   const { data: untrustedAssets } = untrustedAssetsSelector(getState());
   const { data: assetOverrides } = assetOverridesSelector(getState());
+  const { assetOverrides: asssetOverrideSetting } = settingsSelector(
```
quietbits, line 63: Using data from the settings (URL) in some places because the store was causing issues due to race conditions (updating multiple ducks simultaneously or in sequence). It could be better, but it's the best we can do to keep the update small.

Ifropc (approved):

PR #344, opened 2024-05-01
Add additional SEP-9 fields

Ifropc:


packages/demo-wallet-client/src/components/ExtraSep9Fields.tsx
```diff
@@ -48,6 +53,11 @@ const FIELDS = [
     label: "Address",
     type: "text",
   },
+  {
+    id: "customer_id",
```
Ifropc, line 57: This is technically a SEP-24 field, but I assume they are passed in the same manner, so it shouldn't be a problem?
@quietbits WDYT?

quietbits: I think it should be fine. We're not strictly checking against "allowed" SEP-9 fields.

quietbits (approved):

PR #345, opened 2024-05-06
SEP-24: show more_info_url for incomplete transactions

quietbits:
- Show `more_info_url` when we start polling to display it for incomplete transactions.
- Added a `link` prop for the `instruction` log to make a link clickable.


Ifropc (approved):
LGTM!

PR #353, opened 2024-06-27
SEP-6 new flow

quietbits:


packages/demo-wallet-client/src/components/Sep6/Sep6Deposit.tsx
```diff
@@ -360,6 +365,26 @@ export const Sep6Deposit = () => {
...
+        <Modal.Body>
+          <p>Submit the deposit.</p>
+        </Modal.Body>
+
+        <Modal.Footer>
+          <Button onClick={() => dispatch(sep6DepositAction())}>Submit</Button>
+        </Modal.Footer>
+      </Modal>
+    );
+  }
```
quietbits, line 386: We need this extra step to trigger transaction polling after the SEP-12 fields are submitted. We can change the copy of the modal, and the same goes for withdrawal.

philipliu (commented):
LGTM other than the deposit instructions modal.

philipliu (approved):

packages/demo-wallet-client/src/components/Sep6/Sep6Deposit.tsx
philipliu: Will this be added in a different PR?

quietbits: This is already implemented. It looks like it's conflicting with the optional `instructions` for deposit. I'm looking into that.

philipliu: This is an issue with AP's implementation. The transaction moves into the completed status before the wallet polls and sees the `pending_user_transfer_start` status.

JakeUrban:
Are we going to add the SEP-38 functionality in this PR or a subsequent one?

JakeUrban:
@quietbits and I talked offline and decided to merge this PR separately from adding quotes, we'll do that next.

PR #355, opened 2024-07-01
SEP-6 quotes

quietbits:


packages/demo-wallet-client/src/components/Sep6/Sep6Withdraw.tsx
```diff
@@ -253,6 +320,31 @@ export const Sep6Withdraw = () => {
...
+                )}
+                {withdrawResponse.max_amount && (
+                  <p>
+                    <strong>Max Amount: </strong>
+                    {withdrawResponse.max_amount}
+                  </p>
+                )}
+              </div>
+            ) : null}
+          </>
```
quietbits, line 346: We need the amount for the quotes, so I moved it one step back.

philipliu (approved):
I've tested the exchange flows against the Anchor Platform. LGTM!

PR #363, opened 2024-09-05
Fix SEP-6 flow

quietbits:
- Use the correct delivery method
- Show country code even if there is one item, but make it optional (some cases might still require a country code to be sumitted)
- Small cleanup for clarity


philipliu (approved):
LGTM

PR #385, opened 2025-02-21
Fix XLM withdraw failure

lijamie98:
Fix the XLM withdraw submission failure when polling the transaction status.

`new Asset("XLM", "native")` is no longer valid and should be replaced with `Asset.native()`.

### DIFF
--- a/packages/demo-wallet-shared/methods/sep24/pollWithdrawUntilComplete.ts
+++ b/packages/demo-wallet-shared/methods/sep24/pollWithdrawUntilComplete.ts
@@ -12,17 +12,17 @@ import { AnyObject, TransactionStatus } from "../../types/types";
 import { getNetworkConfig } from "../../helpers/getNetworkConfig";
 
 export const pollWithdrawUntilComplete = async ({
-  secretKey,
-  popup,
-  transactionId,
-  token,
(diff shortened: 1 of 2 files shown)


packages/demo-wallet-shared/methods/sep24/pollWithdrawUntilComplete.ts
```diff
@@ -12,17 +12,17 @@ import { AnyObject, TransactionStatus } from "../../types/types";
...
+                                                  popup,
+                                                  transactionId,
+                                                  token,
+                                                  sep24TransferServerUrl,
+                                                  networkPassphrase,
+                                                  networkUrl,
+                                                  assetCode,
+                                                  assetIssuer,
+                                                  sep9Fields,
+                                                }: {
```
JiahuiWho, line 25: Is this change necessary?

JiahuiWho (approved):

packages/demo-wallet-shared/methods/sep24/pollWithdrawUntilComplete.ts
lijamie98: Was from formatter. Reverted for cleanliness.

PR #386, opened 2025-03-06
[Fix] Fix the incorrect interchange of `destinationAssetCode` and `sourceAsset`

JiahuiWho:
This is causing reference server sep6 deposit flow referencing wrong asset in `amountIn` `amountOut` and `amountExpected`


philipliu (approved):

lijamie98 (approved):

JiahuiWho:
In JavaScript, param order in the object doesn't matter, because values were assigned by param name. So changing the order in this fix won't work. Closing this one.

PR #391, opened 2025-05-06
[ANCHOR-1022] Skeleton of create passkey contract using passkey service

JiahuiWho:
This demonstrates the barebones flow for contract account creation via the passkey service.

### What's Changed

C-Account Contract:
1. For now only store passkey id and pk for credential checking.
2. Will be iterated overtime and likely will need to store a asset list that user care about

Passkey service:
1. Only `registerPasskey()` is implemented for now. 
2.`connectPasskey()` will be implemented in next step

SmartWalletService
1. Calling `registerPasskey()`
2. Generate a funded source account
3. deploy contract

UI
1. What's change here are for test purpose only

### Test

1. Using the newest preview link to open demo wallet
2. Click on link "Create new contract account using Passkey (testnet only)"
3. Passkey window will pop out, store your pk in either 1Password, Mac or Chrome etc
4. Open console and wait for ~5 seconds, contract account info will be logged there.

### Limitations

Currently there isn't a way to verify your passkey. It will have to wait until `connectPasskey()` is implemented at the earliest.

PR #394, opened 2025-06-09
[ANCHOR-1027] Fund contract account with XLM

JiahuiWho:
### What's Changed
This PR 
1. Implemented a "transfer" function that enable transfer (9999) XLM from G account to C account upon creation
2. UI improvements including disable all button on landing page when account (either G or C) status is pending
3. Modified `tsconfig` to allow proper sdk import

### Known limitation
9999 XLM is transferred because there's probably some balance check that prevents the G account's resulting balance from being exactly 0 — as shown in the error:
```
[Diagnostic Event] contract:<contract_id>, topics:[error, Error(Contract, #10)], data:["resulting balance is not within the allowed range", 10000000, 0, 9223372036854775807]
```
### Test
1. Click on `Create new contract account using Passkey (testnet only)`
2. Wait for deploy and funding (it takes 10~ seconds), check the contract and balance in stellar expert.

### DIFF
--- a/packages/demo-wallet-client/src/services/SmartWalletService.ts
+++ b/packages/demo-wallet-client/src/services/SmartWalletService.ts
@@ -8,6 +8,7 @@ import {
   nativeToScVal,
   Networks,
   Operation,
+  SigningCallback,
   StrKey,
   Transaction,
   TransactionBuilder,
@@ -26,7 +27,6 @@ import {
   parseRawSimulation,
   Server,
 } from "@stellar/stellar-sdk/rpc";
-import { SigningCallback } from "@stellar/stellar-base";
 
 export class SmartWalletService {
   private static instance: SmartWalletService;
--- a/packages/demo-wallet-client/src/services/SmartWalletService.ts
+++ b/packages/demo-wallet-client/src/services/SmartWalletService.ts
@@ -1,14 +1,28 @@
 import { PasskeyService } from "./PasskeyService";
-import { Address, hash, Keypair, StrKey, xdr } from "@stellar/stellar-sdk";
+import {
+  Address, Asset,
+  hash,
+  Keypair, nativeToScVal, Networks, Operation,
+  StrKey,
+  TransactionBuilder,
+  xdr,
+} from "@stellar/stellar-sdk";
 // @ts-ignore
 import { Client, basicNodeSigner } from "@stellar/stellar-sdk/contract";
-import { SOROBAN_CONFIG, SOURCE_KEYPAIR_SECRET } from "../config/constants";
+import {
+  SOROBAN_CONFIG,
+  SOURCE_KEYPAIR_SECRET,
+} from "../config/constants";
 import base64url from "base64url";
+// @ts-ignore
+import { Api, Server } from "@stellar/stellar-sdk/rpc";
+import { RpcServer } from "@stellar/stellar-sdk/lib/rpc/server";
 
 export class SmartWalletService {
   private static instance: SmartWalletService;
   private passkeyService: PasskeyService;
   private readonly sourceKeypair: Keypair;
+  private rpcClient: RpcServer;
 
   public static getInstance(): SmartWalletService {
     if (!SmartWalletService.instance) {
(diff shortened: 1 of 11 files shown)


packages/demo-wallet-client/src/services/SmartWalletService.ts
```diff
@@ -81,10 +96,98 @@ export class SmartWalletService {
...
+          auth: signedEntries,
+        }),
+      ).build()
+
+    // Step 5: Prepare the transaction
+    const preppedTx = await this.server.prepareTransaction(readyTx)
+    preppedTx.sign(this.sourceKeypair);
+
+    // Step 6: Send the transaction
+    const response = await this.server.sendTransaction(preppedTx);
```
Shaptic, line 166: there's a [`pollTransaction`](https://stellar.github.io/js-stellar-sdk/module-rpc.Server.html#pollTransaction) for this purpose so you can avoid writing your own loop :)

packages/demo-wallet-client/src/services/SmartWalletService.ts
```diff
@@ -81,10 +96,98 @@ export class SmartWalletService {
...
+      throw new Error(simulatedTx.error);
+    }
+
+    // 3. Sign auth entries
+    const signedEntries = await Promise.all(simulatedTx.result?.auth?.map(entry =>
+      authorizeEntry(entry, signer, simulatedTx.latestLedger + 600, Networks.TESTNET)
+    ) ?? []
+    );
+
+    // Step 4: Simulate again to get updated resources
```
Shaptic, line 150: You shouldn't need to do this for a simple transfer :thinking: you should be able to just reassing `simulatedTx.result!.auth` to the `signedEntries` version and call `prepareTransaction`.

yarn.lock
```diff
@@ -1508,39 +1513,40 @@
...
     tweetnacl "^1.0.3"
   optionalDependencies:
-    sodium-native "^4.1.1"
+    sodium-native "^4.3.3"
 
-"@stellar/stellar-sdk@^12.0.0-rc.2":
-  version "12.0.0-rc.2"
-  resolved "https://registry.yarnpkg.com/@stellar/stellar-sdk/-/stellar-sdk-12.0.0-rc.2.tgz#ff467151c825b4a80c9762982d82c7f47f870eeb"
-  integrity sha512-Pj6eBP0doe3yoLa5rhAgDDd+AVSTRKmJFDcnr7LBXt5RzF2P7txpEZ1ZhWG5IggWzJKc/x83zUzOuOfyIglhdA==
+"@stellar/stellar-sdk@13.3.0", "@stellar/stellar-sdk@^13.3.0":
```
Shaptic, line 1540: I'm surprised there's no accompanying `package.json` changes in this PR?

JiahuiWho: `package.json` has been changed in previous PR https://github.com/stellar/stellar-demo-wallet/pull/391/files

philipliu (commented):
Is there a way to show the account as being logged in after the contract account is created? It looks like we are currently waiting for the XLM payment to succeed. Since it's two transactions, we are waiting for 10 seconds without much feedback, which looks like something has gone wrong.

I think ultimately we want to move the XLM transfer into the contract's constructor itself so that the whole flow happens in a single transaction.

Shaptic:
@lijamie98 only if you have > 10k XLM in your account. This error

> 9999 XLM is transferred because there's probably some balance check that prevents the G account's resulting balance from being exactly 0 — as shown in the error

You need at least 0.5 XLM in an account for it to exist (assuming no other data entries in your account like a trustline), this is the [reserve requirement](https://developers.stellar.org/docs/learn/fundamentals/stellar-data-structures/accounts#base-reserves-and-subentries).

JiahuiWho:
> TESTNET actually allows sending 10000 XLMs. Here is an example: https://stellar.expert/explorer/testnet/tx/6309392857174016

It's not about transferring amount limitation. It's about G-account balance can't be 0. The newly created G-account only has 10000 xlm (new account funded with friendbot) so it can only send out 9999. If you have more than 10000 in your G account then of course you can transfer more than 10000, but you can't transfer all of them

philipliu:
> The way to show a c-account is created (or connected) is redirect to the Account Page (https://demo-wallet.stellar.org/account?....)

If it's not possible without redirecting, I think we can fix it later using a single transaction.

PR #419, opened 2025-09-09
Force form data version resolution

JiahuiWho:
`@stellar/stellar-sdk` has a direct dependency on an older version of `axios` that still uses the vulnerable `form-data@4.0.0`, adding `resolutions` to force it resolves to 4.0.4


lijamie98 (approved):
LGTM

PR #458, opened 2026-04-06
Fix `/sign-tx` to allow source account credentials only for contract deployment

JiahuiWho:
#### What's Changed
- [#455](https://github.com/stellar/stellar-demo-wallet/pull/455) rejected all `sorobanCredentialsSourceAccount` auth entries, which blocked both the exploit and legitimate contract deployments. This PR refines the fix to inspect the authorized function: source account credentials are now only permitted for contract deployment (`createContractV2HostFn`) and rejected for everything else.
- Replaced per-route `Access-Control-Allow-Origin` headers with a global Express middleware, ensuring all responses (including errors and preflight OPTIONS) include CORS headers.

#### Context
Error responses were missing CORS headers, causing browsers to mask server errors s CORS failures.

### DIFF
--- a/packages/demo-wallet-server/src/index.ts
+++ b/packages/demo-wallet-server/src/index.ts
@@ -150,9 +150,23 @@ app.post("/sign-tx", async (req, res) => {
       throw new Error("Transaction simulation failed");
     }
     simulatedTx.result?.auth?.forEach((entry) => {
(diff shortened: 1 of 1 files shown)


marwen-abid (approved):
LGTM! Thank you for fixing this.

